PR #1, opened 2015-01-03
Addonify ember-cli

bcardarella:
- [ ] build pipeline
- [ ] blueprint
- [ ] commands
- [ ] test suite

### DIFF
--- a/active/0000-addonify-ember-cli.md
+++ b/active/0000-addonify-ember-cli.md
@@ -1,0 +1,39 @@
+- Start Date: 2015-01-03
+- RFC PR: 
+- ember-cli Issue: 
+
+# Summary
+
+Extract nearly all behavior out of ember-cli core into addons.
+
+# Motivation
+
+I believe that if we can reduce the surface layer API enough for ember-cli it can potentially be a common-purpose build tool for other 
+JavaScript framework. The ember-cli proejct could be a collection of addons that add build-specific tooling for Ember projects.
+The absolute core library would be one that could allow addons to add functionality to the core.
+
+If we can get other frameworks to buy into using certain shared libraries of ember-cli then Ember itself
+benefits.
+
+Another benefit would be opening up alternative implemtation to existing core functionality. While much of 
+what ember-cli does is very good, one could imagine a talented developer not involved with the project creating 
+a competing addon that has significantly better functionality. ember-cli could reliably adopt this upstream if/when it wanted.
+
+# Detailed design
+
+ember-cli is growing in complexity and in some areas getting hairy. Extracting out the behavior into addons will allow
+for more focused development. There could be subject matter expert teams specifically assigned to a given addon. For example
+all of the build pipeline could be an addon (between the app and addons). The behavior in the build pipeline
+has no context relation to say the blueprint system. 
+
+# Drawbacks
+
+It won't be easy. Many of the addon functionality is pretty inter-woven.
+
+# Alternatives
+
+Addons are currently supplementing some built-in behavior but I there isn't much beyond that.
+
+# Unresolved questions
+
+Unknown
--- a/active/0000-addonify-ember-cli.md
+++ b/active/0000-addonify-ember-cli.md
@@ -9,7 +9,7 @@ Extract nearly all behavior out of ember-cli core into addons.
 # Motivation
 
 I believe that if we can reduce the surface layer API enough for ember-cli it can potentially be a common-purpose build tool for other 
-JavaScript framework. The ember-cli proejct could be a collection of addons that add build-specific tooling for Ember projects.
+JavaScript framework. The ember-cli project could be a collection of addons that add build-specific tooling for Ember projects.
 The absolute core library would be one that could allow addons to add functionality to the core.
 
 If we can get other frameworks to buy into using certain shared libraries of ember-cli then Ember itself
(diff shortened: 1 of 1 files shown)


bcardarella:
FWIW if this is a desirable direction I was going to propose this be done in many phases. I can spend time in February breaking out the build pipeline.

jakecraige:
:+1: on this. Should we begin a list on here of specific pieces that should be broken out? A big one that's already on the table is blueprints [#2803](https://github.com/ember-cli/ember-cli/issues/2803)

bcardarella:
@jakecraige added some above, feel free to add/propose more

active/0000-addonify-ember-cli.md
```diff
@@ -0,0 +1,39 @@
...
+# Detailed design
+
+ember-cli is growing in complexity and in some areas getting hairy. Extracting out the behavior into addons will allow
+for more focused development. There could be subject matter expert teams specifically assigned to a given addon. For example
+all of the build pipeline could be an addon (between the app and addons). The behavior in the build pipeline
+has no context relation to say the blueprint system. 
+
+# Drawbacks
+
+It won't be easy. Many of the addon functionality is pretty inter-woven.
```
stefanpenner, line 40: is the idea for these to be inline add-ons or do we rely on npm...

ef4:
I agree with this goal. If we do this it would make sense to create a separate project with a new name to hold the common core infrastructure, and keep ember-cli as the ember-specific consumer of the common infrastructure.

active/0000-addonify-ember-cli.md
bcardarella, line 40: npm

bcardarella:
@ef4 confirm, I would imagine that ember-cli itself would just be a collection of addons that depends upon this more abstract higher-level project

stefanpenner:
I am not apposed to this, but my primary goal will continue to be building the best possible toolchain for ember app developers.

I am fine and happy to enable other efforts, as long as they don't impede on the primary goal of this project.

Ultimately a planned decomposition of the project was a reason behind the org, so lets doit!

bcardarella:
@stefanpenner I can tackle the extraction of the build pipeline in February

stefanpenner:
> @stefanpenner I can tackle the extraction of the build pipeline in February

Awesome, as you begin we should sync up. :)

active/0000-addonify-ember-cli.md
```diff
@@ -0,0 +1,39 @@
...
+- ember-cli Issue: 
+
+# Summary
+
+Extract nearly all behavior out of ember-cli core into addons.
+
+# Motivation
+
+I believe that if we can reduce the surface layer API enough for ember-cli it can potentially be a common-purpose build tool for other 
+JavaScript framework. The ember-cli proejct could be a collection of addons that add build-specific tooling for Ember projects.
```
stefanpenner: typeo proejct

active/0000-addonify-ember-cli.md
```diff
@@ -0,0 +1,39 @@
...
+If we can get other frameworks to buy into using certain shared libraries of ember-cli then Ember itself
+benefits.
+
+Another benefit would be opening up alternative implemtation to existing core functionality. While much of 
+what ember-cli does is very good, one could imagine a talented developer not involved with the project creating 
+a competing addon that has significantly better functionality. ember-cli could reliably adopt this upstream if/when it wanted.
+
+# Detailed design
+
+ember-cli is growing in complexity and in some areas getting hairy. Extracting out the behavior into addons will allow
```
stefanpenner, line 22: this seems more related to the motivation. Lets discuss which pieces specifically should be extracted, and maybe the work that needs to be done to get to those extractions.

This will hopefully allow us to nail some of the high level details, and also share the burden with other contributors.

stefanpenner, line 22: i wonder also, if some of the pieces should be extracted as standalone libraries, rather then ember-addons (or maybe this is your intent). 
- PleasantProgress is a good well tested library, but doesn't need to be an add-on.
- `lib/ext/promise` should become https://github.com/stefanpenner/rsvp-party
-  blueprints should be extracted
- i suspect several parts of ember-cli should become ember-addon's (the testem stuff)

active/0000-addonify-ember-cli.md
stefanpenner, line 40: :+1: 

It also, see some pieces (like the testem stuff) may be better as an addon

active/0000-addonify-ember-cli.md
bcardarella, line 22: @stefanpenner I think both standalone libraries and ember-addons will be the result. There are plenty of candidates for both.

stefanpenner, line 22: let me suggest compiling a table of the potential parts and the intended extraction mechanism?

bcardarella, line 22: @stefanpenner I started to put that into the PR description. Would you prefer it to be in this RFC instead?

stefanpenner, line 22: In this RFC I believe. Ideally someone should be able to read the RFC, and get a decent understanding of the motivations, goal and high level steps to accomplish

bcardarella, line 22: :+1:

trabus:
@jgwhite and I started work on [blueprint extraction](https://github.com/ember-cli/ember-cli/issues/2803) today. The gameplan (as I understand it) is to identify and extract everything that is not ember-cli specific into the blueprint library, then we'll provide the necessary hooks to extend it with what is needed in ember-cli (such as project awareness, ui output, and methods like lookupBlueprint) in an ember-cli-addon.

stefanpenner:
this has gone stale, will gladly re-open if it gets attention.

PR #3, opened 2015-01-10
Ember Doctor RFC

chadhietala:
[RFC](https://github.com/chadhietala/rfcs/blob/ember-doctor/active/0001-ember-doctor.md)

TBD: 
- [x] Expected output
- [x] Allow addons to add health checks and bundled health-checks to be internal addons

### DIFF
--- a/active/0001-ember-doctor.md
+++ b/active/0001-ember-doctor.md
@@ -17,9 +17,9 @@ The motivation behind this is 2 pronged:
 
 The design for this is rather simple. We would first introduce a command called `ember doctor` that would run some default checks. The default checks would do the following:
 
--Run `ember v --verbose` and complain loudly for incompatible versions
--Run `npm outdated --depth 0` to check on outdated modules
--Run `bower list` and display out of date bower components
+- Run `ember v --verbose` and complain loudly for incompatible versions
+- Run `npm outdated --depth 0` to check on outdated modules
+- Run `bower list` and display out of date bower components
 
 These are what is considered default `checks`.
 
--- a/active/0001-ember-doctor.md
+++ b/active/0001-ember-doctor.md
@@ -15,15 +15,13 @@ The motivation behind this is 2 pronged:
 
 # Detailed design
 
-The design for this is rather simple. We would first introduce a command called `ember doctor` that would run some default checks. These default checks would be:
+The design for this is rather simple. We would first introduce a command called `ember doctor` that would run some default checks. The default checks would do the following:
 
-- Node/IO version
-- NPM Version
-- Ember CLI Version
-- Watchman version (if installed)
-- Out of date NPM packages
+-Run `ember v --verbose` and complain loudly for incompatible versions
+-Run `npm outdated --depth 0` to check on outdated modules
+-Run `bower list` and display out of date bower components
 
-In the event that any of these are out of range Ember Doctor should complain loudly.  These are what is considered default `checks`.
+These are what is considered default `checks`.
 
 In your project developers can setup their own Doctor `checks` that get merged in with the default checks. To allow for this Ember CLI will have `ember generate doctor check:service-health`.
 
--- a/active/0001-ember-doctor.md
+++ b/active/0001-ember-doctor.md
@@ -1,0 +1,51 @@
+- Start Date: 2015-01-10
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


active/0001-ember-doctor.md
```diff
@@ -0,0 +1,49 @@
...
+1. Allows developers to submit system level information in pull requests, so that bugs can be filed and potentially replicated.
+2. Gives developers the ability to know about the health of their project and to potentially help with stagnation.
+
+# Detailed design
+
+The design for this is rather simple. We would first introduce a command called `ember doctor` that would run some default checks. The default checks would do the following:
+
+- Run `ember v --verbose` and complain loudly for incompatible versions
+- Run `npm outdated --depth 0` to check on outdated modules
+- Run `bower list` and display out of date bower components
```
stefanpenner, line 22: it may also want to run ember-cli-dependecy-checker

stefanpenner, line 22: - OS information would be great

active/0001-ember-doctor.md
```diff
@@ -0,0 +1,49 @@
...
+These are what is considered default `checks`.
+
+In your project developers can setup their own Doctor `checks` that get merged in with the default checks. To allow for this Ember CLI will have `ember generate doctor check:service-health`.
+
+This command will generate the following directory structure in the root of the project:
+
+```
+doctor/
+  checks/
+    service-health.js
```
stefanpenner, line 34: :+1:

active/0001-ember-doctor.md
```diff
@@ -0,0 +1,49 @@
...
+  index.js
+```
+
+When `ember doctor` is ran we simply will do a merge of the default checks and the ones provided by the application.  
+
+There should also be a way of excluding checks to be ran. Developers should be able to simply pass flags for things they do not care to run e.g. `ember doctor --skip-npm-version`.
```
stefanpenner: i would prefer `--skip ....` that tags multiple arguments, rather then 1 skip flag per check

stefanpenner:
one key point, is the output should be reasonable and concise (if possible)

active/0001-ember-doctor.md
chadhietala, line 22: I was not aware of ember-cli-dependecy-checker till yesterday, I agree it should get ran.  :+1: to OS level stuff.

active/0001-ember-doctor.md
chadhietala: Sounds good.

stefanpenner:
we should also (as a next step) ad some static analysis, to help catch bad practices.

The big one is catching typo's in resolver entry points like `modelFor, store.find, needs etc.`

I suspect this static analysis may be a bit slow (or just harder) to get as part of the build system. Although it would be great if in the future it was part of that, I think adding it to doctor would be great.

My thoughts here are allowing default addons and external addons to add very tasks (including SA) to the doctor.

buschtoens:
@stefanpenner 

> one key point, is the output should be reasonable and concise (if possible)

This could be counter-productive for bug reports. You might be missing just the critical piece of information.

I could imagine something like this:
- Every check is hidden behind a `--flag`, e.g. `ember doctor --dependencies --os`
- Each check can accept options: `ember doctor --dependencies bower,npm --os`  
  Some options are enabled by default. This could be configured in `config/*`.
- `ember doctor` runs a set of default checks with their default options.  
  Could also be configured in `config/*`.
- Checks don't directly output via `console.log` but use a common interface, so the actual output can be abstracted away. Think of console and JSON.

All in all this would enable various levels of verbosity, while providing the full info, if necessary. The JSON file can easily be shared and we can even write a web viewer for those files.

The implementation specifics are not important to me. If you dislike the `--flag` interface, we can easily come up with something different. But I really like the common interface and verbosity options idea.

stefanpenner:
varying levels of verbosity makes sense, but showing someone 100 lines of json when 1 summary line can explain it also doesn't serve anyones purpose.

buschtoens:
That's why we would have sane default options that only display the bare minimum.

stefanpenner:
> That's why we would have sane default options that only display the bare minimum.

+1

abuiles:
@chadhietala added a couple of TBD on the PR based on conv with @stefanpenner

knownasilya:
Can we also add something like `ember doctor prune` which does something along the lines of https://gist.github.com/wrenoud/784dfcff244265f9593b but internally in a more stable way?

PR #4, opened 2015-02-02
Refactor EmberCLI to run Build Pipeline, Express Server & Testem Server in separate processes & corresponding addon API

taras:
[RFC](https://github.com/embersherpa/rfcs/blob/master/active/0000-multiprocess-ember-cli.md)

### DIFF
--- a/active/0000-multiprocess-ember-cli.md
+++ b/active/0000-multiprocess-ember-cli.md
@@ -17,6 +17,8 @@ We provide **serverMiddlware** hook to allow users to add their proxies to the E
 By making it easier to boot multiple processes, we can simplify ember-cli's internal implementation and give a happy path for users to boot their APIs in development.
 
 This refactoring also paves the way for us to simplify deploying to production applications that use Ember Fastboot. By allowing Ember Addons to provide information about processes, we'll be able to generate Profile and (with additional work) proxy configurations that would make deployment to container environments like Dokku & Docker a single step process.
+
+This also replaces the current hacky reloading stuff we do.
  
 # Detailed design
 
@@ -49,11 +51,12 @@ The processes will be automatically stoped when ember-cli is stopped.
 
 # Drawbacks
 
-I don't know.
+- complexity
+- cross platform process management (don't want orphans) *
 
 # Alternatives
 
-Please, suggest.
+From @stefanpenner: stricter internal concept for isolation, or exploration stuff in node 0.11.x and iojs .. I believe @domenic worked on something related, but i maybe be mistaken.
 
 # Unresolved questions
 
--- a/active/0000-multiprocess-ember-cli.md
+++ b/active/0000-multiprocess-ember-cli.md
@@ -1,0 +1,61 @@
+- Start Date: 2015-02-02
+- RFC PR: 
+- ember-cli Issue: 
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


taras:
@bcardarella @stefanpenner @abuiles @rwjblue What do you guys think?

active/0000-multiprocess-ember-cli.md
```diff
@@ -0,0 +1,63 @@
...
+		]
+	}
+}
+```
+
+The processes will be automatically stoped when ember-cli is stopped.
+
+# Drawbacks
+
+I don't know.
```
stefanpenner: - complexity
- cross platform process management (don't want orphans)
  *

active/0000-multiprocess-ember-cli.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Alternatives
+
+Please, suggest.
```
stefanpenner: stricter internal concept for isolation, or exploration stuff in node 0.11.x and iojs .. I believe @domenic worked on something related, but i maybe be mistaken.

active/0000-multiprocess-ember-cli.md
taras: Ok, I will add these 2 points.

I think `node-foreman` might handle the multi platform consideration for us.

stefanpenner: we should investigate how well it works.

stefanpenner:
this also replaces the current hacky reloading stuff we do.

active/0000-multiprocess-ember-cli.md
taras: I can make a prototype and we can try it on windows.

stefanpenner:
I would love to visit this again post 1.0, It will need some investigation but still am +1

stefanpenner:
I like where this is going, but this needs more fleshing out, specifically.
1. error handling (processes fail, either parent or child etc), we need to be sure they crash correctly, and cleanup occurs correctly.
2. what does the process tree look like?
3. ability to reload reprocesses etc.

nathanhammond:
@taras et al.:

The ability to spawn and manage processes as well as pipe information from one process into the next is something that we want inside of ember-cli. This is a required component to be able to build Ember applications and have them play nicely with things like NW.js, Electron, Cordova, and more, many of which have their own specific tools which we need to integrate with gracefully.

For the creating processes themselves, we would like for them to be achievable as an ember-cli addon, as well as being addressable from other "peerDependency" addons. We'd then delegate back to the `ember` invocation at specific points in the lifecycle in order to allow ember-cli to manage the processes.

So, with that, we do want this feature, but we'd like this RFC updated to address specific implementation details for arbitrary process counts, process relationships, bidirectional communication, references to the process itself, failure recovery, and things we've not yet discovered since we haven't tried to complete this design work.

This is part of our "core + extensibility" efforts.

stefanpenner:
> I like where this is going, but this needs more fleshing out, specifically.
> 
> error handling (processes fail, either parent or child etc), we need to be sure they crash correctly, and cleanup occurs correctly.
> what does the process tree look like?
> ability to reload reprocesses etc.

^^ should be addressed for this RFC to help us explore the solution here.

Turbo87:
closing due to inactivity 😢

PR #5, opened 2015-02-09
Structures: A new model to represent an asset organization structure and resolver rules to find those assets.

trabus:
[RFC](https://github.com/trabus/rfcs/blob/e68c778593b8ad924893cf9e8d6f730b250451bf/active/0000-structures.md)

This is sort of in progress still, because there are lots of moving parts, and some of the ideas around the concept that are still solidifying for me. I wanted to get my first draft up for review, but I wanted to note that I will likely make some revisions within the next couple of days.


jgwhite:
I really like where this is headed. It feels like a natural companion to the blueprints system.

We have seen reasonable evidence of folks in the community using bespoke variations on the pod scheme. Allowing users to declare their project’s structure as a first class API seems like a clear win.

As for packaging the resolver with the structure — I think this warrants more bikeshedding. Would love to hear from folks who are doing custom stuff in their apps today.

jgwhite:
As evidenced by: https://github.com/ember-cli/ember-cli/issues/3234

chadhietala:
Yes to this, I'm running up to the point were neither pods nor basic layout is ideal.  I actually wrote a custom resolver last night and the realized I was going to have to implement all of the generators :-1:. So yea this would be great.

stefanpenner:
this is interesting, but let me play devils advocate. Is there a file-layout that would actually satisfy the current deficiencies and do so in a way that all ember-apps have a similar look and feel? I am extremely nervous to encourage arbitrary structures, as we do want some semblance of consistency between Apps.

That being said, I suspect internally a structure could implement ^^ in a lovely way.

/end brain dump

trabus:
> I am extremely nervous to encourage arbitrary structures, as we do want some semblance of consistency between Apps.

I think we'd probably need to heavily warn against using them to implement arbitrary structures. I'd even be willing to go as far as to limit them to internal use only. But the ability to modularize the way ember-cli is aware of it's structures it can generate blueprints and lookup modules in would be really powerful. It would give us the ability to easily experiment and extend the structures ember-cli supports.

stefanpenner:
> I think we'd probably need to heavily warn against using them to implement arbitrary structures. I'd even be willing to go as far as to limit them to internal use only. But the ability to modularize the way ember-cli is aware of it's structures it can generate blueprints and lookup modules in would be really powerful. It would give us the ability to easily experiment and extend the structures ember-cli supports.

I'm glad we are on the same page.

Should this RFC include some structures we may want to experiment with? That way the proposed API can "theoretically" be validated for completeness.

trabus:
> Should this RFC include some structures we may want to experiment with? That way the proposed API can "theoretically" be validated for completeness.

Sure, we could look at some of the alternative pod structures recommended [here](https://github.com/ember-cli/ember-cli/issues/3234), and slabs/bundles (or whatever they end up being called).

DougPuchalski:
Personally I experience a significant productivity hit when I am forced to name files as their type rather than something more meaningful, despite the implicit directory structure. Depending on your editor, IDE, or overall workflow I hope the impact of this isn't ignored.

If there's a new structure, perhaps it can be flexible with ultimate file naming, i.e. `/components/my-component/my-component-template.hbs` works as well as `/components/my-component/template.hbs`.

MiguelMadero:
We have a large-ish Ember Application where we have multiple domains. We have one top level folder for each of these domains, internally the domains use either the pods or the classical structure. In ember-cli we don't have an easy way of saying, "create a component and by the way that belongs to the `labs` area" or the "clinical" area. There's a big overlap here engines, but once it comes to structure that's the way we're currently organizing our code.

Eptis:
@aceofspades I think this would be nice as well, since I know have to type in sublime `<resource-name>/template.hbs` and stuff, which is less efficient than immeadiately being able to type `<resource-name>.hbs`

active/0000-structures.md
```diff
@@ -0,0 +1,150 @@
...
+* Token logic
+  * Provide token names/values used for fileMapTokens
+* isSupported method
+  * Runs a series of checks to see if a passed in blueprint can support the structure
+* Resolver lookup pattern/rules
+  * (if any, the basic structure wouldn't need any)
+
+
+### Rough Concept
+Structures would be contained in a folder on the root of the ember-cli project, much like the blueprints folder. Structures should be node modules, named by their structure name, and only contain an object to be used in the creation of a new Structure instance:
```
stefanpenner, line 42: what about add-on provided structures?

active/0000-structures.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+
+### Steps to implement
+After the Structure class has been completed, the first step of this would probably be to extract pod lookup patterns from the resolver into a structure and implement them in ember-cli
+
+
+# Drawbacks
+
+It may be a overly-complicated solution for a problem that we don't have many use cases for yet.
```
stefanpenner, line 142: Yes I agree, although it may be a nice internal refactoring. We should only proceed if this ends up feeling like a net benefit. I believe this is absolutely possible though.

active/0000-structures.md
```diff
@@ -0,0 +1,150 @@
...
+
+# Alternatives
+
+This could be built into blueprints directly as an object without the need for a stand alone class.
```
stefanpenner, line 146: potentially, i would love to here where they are similar, or different.

I personally prefer multiple collaborators with extremely clear interactions of mega objects.

active/0000-structures.md
trabus, line 142: I'm aiming to start work on this beginning of June once the dust has settled from my move. I feel this will be an important part of making `ember-mv` easier to implement.

stefanpenner, line 142: :+1:

rwjblue:

Thank you!

PR #6, opened 2015-03-05
Extendable Commands

davewasmer:
[Rendered](https://github.com/davewasmer/rfcs/blob/master/active/0000-extensible-commands.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,9 @@ alternatives tend to be poorly-received**.
 * Submit a pull request. As a pull request the RFC will receive design
 feedback from the larger community, and the author should be prepared
 to revise it in response.
+* Find a champion on the [ember-cli core team](http://emberjs.com/team/). 
+Without a champion, the RFC will not progress. This champion will be assigned
+the RFC in question.
 * Build consensus and integrate feedback. RFCs that have broad support
 are much more likely to make progress than those that don't receive any
 comments.


davewasmer:
@rwjblue what's the process for these RFCs? Should I work on an actual PR for this, or should I try to gather feedback on this RFC first? What are the next steps for merging this PR in?

active/0000-extensible-commands.md
```diff
@@ -0,0 +1,62 @@
...
+* Extending built-in commands could be a bit of a footgun - it would allow the addon developer to deviate from the conventions that this project attempts to cultivate.
+* Using the implementation approach outlined above, the addon that fires first for a built-in command could forgot to call super, causing a break in the chain and the default functionality not to run.
+
+# Alternatives
+
+* Rather than allowing complete overrides of commands, we could just provide hooks for common operations (i.e. modify the `availableOptions` hash, `anonymousOptions` array, etc). This would help prevent abuse, but limits use cases.
+* Even more restrictive, we could not allow any addon modification, and instead wait for common patterns or use cases to arise, and bake those into the built-in commands via optional arguments, auto-detecting addons, config, etc. This is difficult over the long term though, and doesn't scale well.
+* Finally, we may not want to allow customization of built-in commands at all, if the potential for abuse and surprising behavior outweighs the benefits of customization.
```
jakecraige: I'm in favor of this option. Extending the defaults into a custom named command seems like it covers the case of someone wanting to do this well, while preventing user confusion.

Having to type `ember custom:test` vs. `ember test` doesn't seem like to much to ask. It also aligns well with Ember 2.0 of taking away magic by making things explicit.

I'm happy to hear other's thoughts on this.

davewasmer: I would disagree (no surprise there :wink:).

I view the CLI commands as the useful abstractions whose implementation details can vary. In the same way that I should be able to use `store.find('book')` without knowing exactly how the adapter fetches those books.

The intent is clear when I run `ember test`, and if I'm a new dev coming on to an existing Ember project, it makes it much simpler to onboard thanks to the conventional approach. I don't have to learn _this_ project's test command, generate command, etc, I just use the conventional commands, which may have been tweaked by the unique needs of this project, but still carry the same intent.

All that said, I understand the need to take away the magic and make things explicit. Perhaps there is a way it could be obvious when a built-in command is modified. Maybe the command output could show the stack as it's executing, so you get a sense of `super-test-addon:test` running, then built in `ember-cli:test`, etc.

jakecraige: That's an interesting view of the commands, maybe it's a new ember-cli thing, but I've never viewed commands on the command line as abstractions. Perhaps I need to expand my mind, lol :) 

Depending on other's thoughts, I think we could definitely come up with a way to provide context if we decide to allow people to override commands. Currently we allow generators to override the defaults which is a similar case. When you run `ember g -h` it outputs it in the section for the addon that provided it, so we could do the same thing with `ember -h`. It could be pretty annoying if every time you ran the overridden command it bothered you about it being from an addon

rwjblue: I generally agree with @davewasmer on this, but definitely think we need to:
1. Show at invocation time (possibly only with a `-v` or `--verbose` flag?) that the default command is augmented.
2. Make an API internally that makes it possible to "disable" extension.  That way we could ensure things like `ember install:addon`, `ember init`,  etc are safe and cannot be extended...

davewasmer: @rwjblue:
1. I like the `-v/--verbose` flag. I'll add that in. **Edit:** [added](https://github.com/ember-cli/rfcs/pull/6/files#diff-e0dc7808988858b798f17aa1236e242dR48)
2. Are you saying there are certain commands that _cannot_ be extended, or are you saying there should be some API to run certain commands with their extensions disabled? I personally would favor the latter, but could be persuaded otherwise.

rwjblue: > Are you saying there are certain commands that cannot be extended, or are you saying there should be some API to run certain commands with their extensions disabled?

I think we can have both actually. My suggestion is an API to disable extension/modification (the first option) that would be used by Ember CLI to implement the second option. Something like:

```
// default implementation
module.exports = Command.extend({
  allowOverrides: function() {
    var someVarThatMeansTheUserHasDisabledOverrides = "???";

    return !someVarThatMeansTheUserHasDisabledOverrides
  }
});
```

active/0000-extensible-commands.md
```diff
@@ -0,0 +1,64 @@
...
+invokeNextCommandInStack()
+```
+
+> **Note:** obviously, there's a lot more that should happen, and the commands are not functions themselves. Likely, either the Command model would need to incorporate the idea of a "stack", the `/lib/cli/cli.js` file would have to become stack-aware, or we introduce a new CommandStack abstraction to handle this.
+
+We would also add a `-v/--verbose` flag which would print out information about what extensions are running. Output could show what addon in the "stack" is currently executing, which would make it easier to track down what code could be introducing a bug.
+
+# Drawbacks
+
+* Extending built-in commands could result in user confusion, i.e. I checkout an Ember CLI project and run `ember test` and it performs unexpected behaviors, so I Google around for "ember test" and none of the standard docs help me.
```
stefanpenner: i think this is ok though and is expected behavior i believe.

stefanpenner: but I wonder if it should just be 100% replacement, not inheritance merge. Inheritance here is tricky, i suspect it will just get sloppy.

davewasmer: I think 100% replacement plus a good set of hooks is probably a good solution (like how addons currently have hooks for various build steps)

stefanpenner:
cc @jgwhite @trabus  (just making sure you are aware of this one)

trabus:
I'm :+1:  on the concept, but uncertain on the best way to implement safely. We definitely need to be careful in the implementation.

Also I just wanted to point out that the blueprint model makes use of the [`Command.registerOptions`](https://github.com/ember-cli/ember-cli/blob/master/lib/models/command.js#L71) method, which essentially achieves what this RFC calls for, but for a specific blueprint instead of a command.

active/0000-extensible-commands.md
stefanpenner: Ya, although tricky to implement as we move these commands forward.

An alternative approach is `ember test:mocha` etc, but that also feels poor. I would love to see someone explore deeply what `ember test` `ember test --server would exactly need to do, to work nicely with both qunit/mocha/etc while allowing the add-ons to provide the needed details.

active/0000-extensible-commands.md
```diff
@@ -0,0 +1,64 @@
...
+* Extending built-in commands could be a bit of a footgun - it would allow the addon developer to deviate from the conventions that this project attempts to cultivate.
```
stefanpenner: So, maybe we are thinking about this too generally. It seems like the `ember test` and `ember test --server` scenarios are the ones  this RFC details. Maybe they just need to be massaged further?

stefanpenner:
I think we are still lacking here, I like where this is going. But I feel like we need to actually figure out the specifics (non general solution) for a-few variants first, before we can extract some general rule. I don't feel comfortable guessing at the general form quite yet.

stefanpenner:
In the world of extending, would one just replace a command, inherit from the provided one and then augment as needed? Is it that we actually need to define the public interface of the common commands, so it is more obvious what needs to be done?

PR #7, opened 2015-03-24
Addon Options

knownasilya:
Similar to what Yeoman does.

[Rendered](https://github.com/knownasilya/rfcs-1/blob/addon-options/active/0000-addon-options.md)

---

Current implementation is happening in https://github.com/ember-cli/ember-cli/pull/5708

### DIFF
--- a/active/0000-addon-options.md
+++ b/active/0000-addon-options.md
@@ -1,0 +1,33 @@
+- Start Date: 2015-03-24
+- RFC PR: 
+- ember-cli Issue: 
+
+# Summary
+
+Doing something similar to what Yeoman does would be awesome.
+
+![from yeoman](http://yeoman.io/assets/img/codelab/image_8.5a17.png)
+
+Basically exposing some of those cli attributes as questions, and being able to add additional addons/settings/files based on user's answers.
+
+Like Less, Bootstrap, FontAwesome, etc..
+
+# Motivation
+
+This would also make addons much more powerful, since they can ask the user critical questions, and not always assume. This would also make ember-cli more powerful, and as we draw closer to a core cli, these features will be more important for wider adoption.
+
+# Detailed design
+
+TODO
+
+# Drawbacks
+
+More API surface area.
+
+# Alternatives
+
+None.
+
+# Unresolved questions
+
+How to do this.
--- a/active/0000-addon-options.md
+++ b/active/0000-addon-options.md
@@ -42,9 +42,7 @@ availableOptions: [{
 }]
 ```
 
-A new CLI flag is going to be added, called `--skip-prompts` which basically doesn't run the prompts portion of the available options. This flag, along with the `this.ui.ci` are used to skip the prompts explicitly or implicitly if running the addon
-in a CI environment, like Travis. This allows for a graceful fallback, and also a way to not break user's tests or require
-user feedback for whatever reason, like during debugging.
+A new CLI flag is going to be added, called `--skip-prompts` which basically doesn't run the prompts portion of the available options. This flag can be used to skip the prompts explicitly if running the addon in a CI environment, like Travis. This allows for a graceful fallback, and not breaking the user's tests or require user feedback for whatever reason, like during debugging.
 
 An addon developer doesn't need to define all of the inquirer.js prompt options, since we use some defaults
 from the parent attributes, like `name`, `description` and `type`. Although, we still allow for fine-grained configuration
--- a/active/0000-addon-options.md
+++ b/active/0000-addon-options.md
@@ -1,6 +1,6 @@
 - Start Date: 2015-03-24
-- RFC PR: 
-- ember-cli Issue: 
+- RFC PR:
+- ember-cli Issue:
 
 # Summary
 
@@ -18,11 +18,41 @@ This would also make addons much more powerful, since they can ask the user crit
 
 # Detailed design
 
-**TODO**
+Prompts are defined at the blueprint level, so you can create
+a prompt that runs during `ember install my-addon` or during the explicit `ember g my-blueprint`.
 
-Add a `prompts` hook for addons, which would utilize `ui.prompt` behind the scenes.
+Prompts are defined using `ui.prompt`, which pulls data from the blueprint's `availableOptions` if it has `argName` defined or there are options available. For example:
 
-One issues that prompts introduce is when you aren't interacting with your setup i.e. CI or testing, and also the possibility to overwhelm a user. Having a `--skip-prompts` option would mitigate issues like that, and prompts should have defaults, so skipping prompts would just do the default.
+```js
+availableOptions: [
+  name: 'local',
+  type: Boolean,
+  aliases: [ 'l' ],
+  default: false,
+  description: 'whether release commit and tags are locally or not (not pushed to a remote)',
+  validInConfig: true
+]
+```
+
+Which can be utilized by `ui.prompt` automatically as long as the author does something like:
+
+```js
+return this.ui.prompt([
+  {
+    name: 'useLocalGit',
+    type: 'confirm',
+    values: ...,
+    argName: 'local' // this maps to the option `name` above
+  }
+  // more prompts
+]);
+```
+
+This way the prompt can pull in default values or values specified via the CL, e.g. `--local`. Default values would be used if `--skip-prompts` is used.
+
+As a fail safe, addon authors will have a warning or an error if there are no default values for their prompts. This ensures that addons can be used in tests and CI systems, as well as for user convenience.
+
+Prompts can be specified in `beforeInstall`, `beforeUninstall`, `locals`, `afterInstall` and `afterUninstall`. It should always be used in the promise form, i.e. no callback.
 
 # Drawbacks
 
@@ -35,4 +65,4 @@ None.
 
 # Unresolved questions
 
-How to do this.
+* `ember install` can take a list of addons. This creates a problem of possible collisions for CL arguments mapping to prompts. Possible solution is to pass all args and hope for the best, or the alternative would be to allow for a prefix.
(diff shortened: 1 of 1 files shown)


davewasmer:
Isn't this available via `myAddon.ui.prompt`?

active/0000-addon-options.md
```diff
@@ -0,0 +1,33 @@
+- Start Date: 2015-03-24
+- RFC PR: 
+- ember-cli Issue: 
+
+# Summary
+
+Doing something similar to what Yeoman does would be awesome.
```
stefanpenner: please actually elaborate, these RFC's are meant to flesh out the details so implementation is clear.

knownasilya: Updated.

knownasilya:
@davewasmer yes it looks like inquirer.js is exposed, but I think we need an addon hook for this, with the ability to specify generators by name for yes/no options, etc. Still thinking on the API/design, feedback welcome!

stefanpenner:
during what phase of an addons life-cycle should this prompt appear. I am apposed to this happening during the install phase, if an add-on author would like to add a command or special purpose generator. I think that is alright.

for example

`ember deploy:divshot` or `ember g divhsot:setup` I believe could reasonably prompt the user for information here.

`ember install ember-divshot` I don't feel comfortable with this prompting, as it would create a pretty horrible experience if 20 addons all pause the installation process to prompt the user.

knownasilya:
@stefanpenner having a `--skip-prompts` option would mitigate issues like that, and prompts should have defaults, so skipping would just do the default.

knownasilya:
Having prompts at install is probably one of the most useful features, since most work is done there, hence the default generator for `ember install:addon`.

stefanpenner:
Actually You may be correct about this.  I'll enumerare my hesitations later this mornjnf

active/0000-addon-options.md
```diff
@@ -0,0 +1,38 @@
...
+
+Enabling addons to specify custom prompts that would inquire the user, and perform arbitrary actions based on those responses. Actions could range from running blueprints, installing additional addons, setting configuration.
+
+Similar to what Yeoman has:
+
+![from yeoman](http://yeoman.io/assets/img/codelab/image_8.5a17.png)
+
+# Motivation
+
+This would pave the way for users creating custom 'app' blueprints that can have their personal/comapany's workflows, with default settings, and optionals. This could include installing the addons you always install, e.g. ember-cli-less, bootstrap, font-awesome, etc.
```
stefanpenner, line 15: I am fine with:
- custom blueprints having prompts
- non-default generator steps having prompts
- custom commands having prompts

for example:

```
ember emberfire:setup
ember g some-cool-generator-that-asks-a-question
ember new --blueprint my-blueprint name
```

I am concerned with the default generator issuing prompts, as I feel it is extremely poor UX to get a barrage of questions you may not be prepared for when installing several add-ons. It can also be non-obvious how to `undo`.

In-theory, if we extended our undo/apply pattern to these we could offer a better experience. But that sounds extremely tricky and I am unsure without further investigation

knownasilya, line 15: Agreed. I've tried maintaining an 'app' blueprint, and it's an uphill battle since EmberCLI changes so quickly.

stefanpenner, line 15: @knownasilya ya, we need to improve inheriting/extending/configuring blueprints from addons.

I believe https://github.com/ember-cli/rfcs/pull/8 is a small step in the correct path, especially with: https://github.com/ember-cli/rfcs/pull/8#discussion_r30533585  thoughts?

knownasilya, line 15: Agreed. I was actually thinking of doing something similar with prompts, where a custom prompt exists that installs packages with the same kind of format, i.e. a list of packages.

So a single prompt could install a single package or several.

> Would you like to setup the default UI libraries?

Something along the lines of (not covering the bower/npm use case -- not sure I want too lol).

``` js
{
 type: 'packages',
 message: 'Would you like to setup the default UI libraries?',
 packages: [
   'fontawesome',
   'bootstrap',
   'ember-cli-less'
 ]
}
```

stefanpenner:
@runspired i believe this has some overlap with your work. I would love to see us flesh this out, evolve it and implement something once its ready.

active/0000-addon-options.md
```diff
@@ -0,0 +1,38 @@
...
+
+This would also make addons much more powerful, since they can ask the user critical questions, and not always assume. This would also make ember-cli more powerful, and as we draw closer to a core cli, these features will be more important for wider adoption.
+
+# Detailed design
+
+**TODO**
```
stefanpenner: this section needs to be fleshed out

knownasilya: Yes please. We need to solve https://github.com/ember-cli/ember-cli/pull/4025 first I think.

I think we should basically expose a working prompts interface, and work from there based on usage in addons. Abstracting as we see patterns.

runspired:
This is right up my alley, I do a lot with prompting during installation of the default blueprint, it's an ideal time to ask the user to configure additional things to be installed.  For instance, installing ember-bootstrap might be segmented into "just css" "also fonts" "also js components" and the like.  Many addons use this time to prompt the user for configuration of whether they would like to import the addon's default CSS.

runspired:
@knownasilya @stefanpenner the question I have here is if this isn't an unnecessary complication.  As it stands, at least `beforeInstall` and `afterInstall` are promise aware and have access to `this.ui.prompt`, what does the prompts hook gain us here?

knownasilya:
@runspired I wasn't aware of those hooks. Maybe those hooks along with the `locals` hook would be enough? I guess the only thing left to do is make it work and document it.

stefanpenner:
I believe we need a real example in the motivations sections, otherwise abstractly its hard to agree/disagree or know for-sure.

knownasilya:
@stefanpenner mind having a look at the updates?

active/0000-addon-options.md
```diff
@@ -0,0 +1,81 @@
...
+# Motivation
+
+This would pave the way for users creating custom 'app' blueprints that can have their personal/comapany's workflows, with default settings, and optionals. This could include installing the addons you always install, e.g. ember-cli-less, bootstrap, font-awesome, etc.
+
+This would also make addons much more powerful, since they can ask the user critical questions, and not always assume. This would also make ember-cli more powerful, and as we draw closer to a core cli, these features will be more important for wider adoption.
+
+# Detailed design
+
+Prompts are defined at the blueprint level, so you can create
+a prompt that runs during `ember install my-addon` or during the explicit `ember g my-blueprint`.
```
stefanpenner, line 22: I do not believe we should support prompts during the installation or default generator phases. This is extremely disruptive. It may be possible for us to allow users to opt-into an interactive mode during the initial install / blueprint phase, but I'm :-1: on it being by default.

Addon's should provide sensible defaults. Secondary generators are fine to use a prompt.

active/0000-addon-options.md
```diff
@@ -0,0 +1,81 @@
...
+For advanced use-cases the addon author can dig into the prompt results and tweak using the returned promise.
+
+```js
+return this.ui.prompt([
+  // prompts
+]).then(function (answers) {
+  // tweak the answers
+  return answers;
+})
+```
```
stefanpenner, line 58: what happens if the user `cancels`/`escapes` the prompt, via `<ctrl-c>` or similar.

active/0000-addon-options.md
knownasilya, line 22: See https://github.com/ember-cli/rfcs/pull/7#issuecomment-85587635.

What about `ember install my-addon --interactive`? But then you need to know that the addon is interactive, maybe we can make it an option that can be set in the `.ember-cli` file for those that want it all the time.

active/0000-addon-options.md
knownasilya, line 58: I think it'd cancel whatever command you were doing,  and you have to try again..

active/0000-addon-options.md
knownasilya, line 22: @stefanpenner thoughts on this last comment?

stefanpenner, line 22: @knownasilya sorry for the very late response `--interactive` sounds like a great compromise, so by default we will run with `--interactive=false` and individuals could opt-into `--interaction=true`, or even via emberrc if they always wanted it.

This seems good, as it continues to allow a default happy-paths without going down the choose your own adventure path, but for those who prefer it, they could easily opt-in.

active/0000-addon-options.md
```diff
@@ -0,0 +1,76 @@
...
+    //name - defaults to camelized name above
+    //type - defaults to logical mapping from the same field above
+    //default - defaults to the value above
+    //message - defaults to the description above
+    //..any other options from inquirer.js
+  }
+}]
+```
+
+A new CLI flag is going to be added, called `--skip-prompts` which basically doesn't run the prompts portion of the available options. This flag can be used to skip the prompts explicitly if running the addon in a CI environment, like Travis. This allows for a graceful fallback, and not breaking the user's tests or require user feedback for whatever reason, like during debugging.
```
stefanpenner, line 45: this is replaced by `--interactive=<true|false>` right?

active/0000-addon-options.md
```diff
@@ -0,0 +1,76 @@
...
+* More API surface area (more to test, and fix).
+
+
+# Alternatives
+
+None.
+
+# Unresolved questions
+
+* `ember install` can take a list of addons. This creates a problem of possible collisions for CL arguments mapping to prompts. Possible solution is to pass all args and hope for the best, or the alternative would be to allow for a prefix.
```
stefanpenner, line 75: 1. one addon, pass args
2. > one addons, either pass all args or none (seems like none make sense?), if they wish they can do `--interactive=true`

active/0000-addon-options.md
knownasilya, line 45: See last discussion here https://github.com/ember-cli/ember-cli/pull/5708#discussion_r58053107

knownasilya, line 45: Ping @stefanpenner

active/0000-addon-options.md
knownasilya, line 75: So basically the current implementation makes sense. Addon namespaces are a bigger concern and not part of this discussion.

rwjblue:

Thank you!

PR #8, opened 2015-03-27
Optable-features in app blueprint - blueprints

abuiles:


stefanpenner:
I would prefer to ensure blueprints support the various needed features to enable this, adding a wide-array of flags and switches to cover this functionality seems like an anti-pattern.

abuiles:
@stefanpenner

> adding a wide-array of flags and switches to cover this functionality seems like an anti-pattern.

You are right. I checked and we don't need to make this options part
of the command since we can specify options directly in the
blueprints, which led us to

> I would prefer to ensure blueprints support the various needed
>   features to enable this

I think we have most of the hooks to enable this, for the initial scenarios we might need:
- [ ] Discard files from `/files`: `_processBlueprint` hook.
- [ ] Edit files after they have been installed: `afterInstall` hook.

cc @jgwhite

stefanpenner:
> Edit files after they have been installed: afterInstall hook.

this would likely break the transactional model our blueprints currently have, rather allowing for AST transforms during the blueprint (after the template step) phase via babel propose 5.x transform API, seems like a good option for this.

My feeling is nearly all of these concerns should be addressable via blueprints. If they aren't we should fix those issues.  I suspect we should also make blueprints easier with better documentation and more examples.

I do also suspect some stuff may fall outside of the blueprint concept, but we should investigate on a case by case basis.

abuiles:
> this would likely break the transactional model our blueprints
> currently have

We already have some blueprints which do this, for example the route
and addon blueprints.

> rather allowing for AST transforms during the blueprint (after the
> template step) phase via babel propose 5.x transform API, seems like a
> good option for this.

Sounds good, so if `this.transforms(file, AST)`, then yield an AST for
every file, should we write back automatically or let that to the
consumer?

stefanpenner:
We should transform inflight as the file info is applied

abuiles:
> We should transform inflight as the file info is applied

sounds good, I'll add  this to the design section of the RFC. Also, I might rename it again.

stefanpenner:
> sounds good, I'll add this to the design section of the RFC. Also, I might rename it again.
>  Merge pull request

awesome, ya lets evolve this one to handle the motivating concerns

benlesh:
Would there be a master blueprint that ran these on `init` or `new`? Some of these features might be things the majority of people want, I guess. Also: Would the initialization of the git repo be covered by this as well? Not everyone uses git or can use git. I've been forced to use TFS in the past (shudder).

active/0000-optable-features.md
```diff
@@ -0,0 +1,50 @@
...
+People can save their default in `.ember-cli` and they won't need to
+memorize them every time they run `new` or `init`.
+
+Additionally we could introduce the concept of a ".global-ember-cli"
+where people can specify default options for all their ember-cli
+projects.
+
+# Drawbacks
+
+TBD
```
stefanpenner, line 41: inner complexity

we need to be sure:
- internals improve to handle this
- docs are clear
- hooks for add-ons are clear.

but i am +1 on this.

active/0000-optable-features.md
```diff
@@ -0,0 +1,50 @@
...
+
+# Alternatives
+
+At the moment it seems like this can be done at some level with
+blueprints, but I haven't confirmed yet.
+
+# Unresolved questions
+
+Full list of initial options.
```
stefanpenner, line 50: lets flesh this list out, although the above seem good as a first start.

active/0000-optable-features.md
```diff
@@ -0,0 +1,50 @@
...
+- `--stylesheet sass`
+- `--template emblem`
+- `--testing karma`
+
+People can save their default in `.ember-cli` and they won't need to
+memorize them every time they run `new` or `init`.
+
+Additionally we could introduce the concept of a ".global-ember-cli"
+where people can specify default options for all their ember-cli
+projects.
```
stefanpenner, line 37: blueprints should be able to tap into this

for example

```
ember new --blueprint https://github.com/stefanpenner/my-blueprint my-app
```

may just be a bag of options applied to the existing app blueprint. This would by itself satisfy many of my own needs for custom blueprints.

We would likely need to expose --add-packages aswell, likely using the npm micro-syntax (lets ignore the existence of bower here).

for example:

```
ember new --add-packages=rsvp@3.0.14 my-cool-addon yahoo/ember-app@^2.0.0
```

now if these could be defined in the blueprint, this would be rad.

``` js
module.exports = {
  addPackages: function() {
    return [
      'rsvp@^3.0.0',
      'my-cool-addon',
      'yahoo/ember-app@^2.0.0'
    ]
  }
}
```

stefanpenner:
I have been noodling on blueprint improvements for some time now, I believe some refactoring of the blueprints needs to land first, before we should implement this, as I believe it can be simplified and many long lasting issues addressed.

Abstractly I feel blueprinting should be 3 parts:
1. a series of operations.
2. a reduction of those operations.
3. application of those operations.

The goal of this, is to allow blueprint non-brittle blueprint composition

given a set of operations:

```
[file:create app/app.js]
[file:transform, editorconfig, **/*.js] // imagine, a AST transform to update blueprints to match editor config
[file:transform, prependHeader, **/*.js] // insert a license header
[ember:install
[file:create app/app.js] // from an alternative blueprint, maybe the user provides different defaults
[ember:install some-other-module]
```

Operations can specify several things:
1. which file they are associated too.
2. if they are "unique", file:create is unique, and how to handle dedupes (in this case the last one is taken, in other cases maybe they error, or the first one is taken).
3. what dependencies they have, some transformations may be order-dependent
4. ... ?

So given the above example, we will want the following reduction.

```
[file:create app/app.js] // we choose the last one. (allowing the user-land blueprint to provide a different default)
[file:transform, editorconfig, **/*.js] // transforms are after file-creations
[file:transform, prependHeader, **/*.js] 
[ember:install, some-module, some-other-module] // are combined, and played at the end.
```

This allows us to discover all required actual work, then prompt the user on conflict, once all the work has been confirmed we can apply the changes.

### Other thoughts
- Custom operations, allowing for more arbitrary work. For example, imagine a operation to upgrade/downgrade/install/remove various dependencies IFF some precondition is met.

---

I will follow up with a more detail RFC on this topic at some point. Hopefully soon.

trabus:
@stefanpenner :+1: on your thoughts.
In addition to fleshing out the whole order of operations, `file-info.js` will need some refactoring to accommodate a way to manage and apply operations to the files. It would likely be a good place to queue, reduce, and resolve operations for each file.

stefanpenner:
> It would likely be a good place to queue, reduce, and resolve operations for each file.

ya, i think it could store all operations on it, the once that flush is complete, we can reduce and then prompt the user.

stefanpenner:
Im on-board with improving this, but it needs to be actively championed. If there is interest, and a champion we can gladly reopen.

PR #11, opened 2015-04-26
Generate app from template

bcardarella:


bcardarella:
If this passes muster I'm willing to put a **$500 bounty*** on this feature to ensure this gets in prior to `1.0`.

<sup>*Does not apply to DockYarders</sup>

rwjblue:
I believe this is already implemented:

```
ember new --blueprint=my_blueprint
```
- A [few tests showing usage](https://github.com/ember-cli/ember-cli/blob/a440c4ee1b0f9617704228f91e14b5f2b41ba75f/tests/acceptance/new-test.js#L115-L171).
- Implemented/touched in the following PR's:
  - https://github.com/ember-cli/ember-cli/pull/462
  - https://github.com/ember-cli/ember-cli/pull/2103

---

There is still a but of work if you want this to work from an internet hosted `.tar.gz` file though (I don't believe that is implemented yet), but it could also be relatively easily wrapped in a two line shell script or something...

bcardarella:
@rwjblue my understanding is that the current functionality requires you to maintain the entire blueprint set somewhere. I don't think this is a good approach. Instead there should be a simple set of rules for changes that should be made after the app is generated normally.

rwjblue:
You would invoke the main `app` blueprint, then make the smallish modifications you need in your custom blueprint. The pattern of invoking other blueprints from within a blueprint is one that is done often in the codebase already (`ember g resource` generates routes, controllers, templates, etc).

bcardarella:
This works!

PR #12, opened 2015-05-16
Add option for JSON output to help command

thomassnielsen:
[Rendered markdown document](https://github.com/thomassnielsen/rfcs/blob/help-json-output/active/0000-help-json-output.md)

### DIFF
--- a/active/0000-help-json-output.md
+++ b/active/0000-help-json-output.md
@@ -1,0 +1,69 @@
+- Start Date: 2015-05-16
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+This has come up in [#3699](https://github.com/ember-cli/ember-cli/issues/3699) and [EmberTown/ember-hearth/#66](https://github.com/EmberTown/ember-hearth/issues/66).
+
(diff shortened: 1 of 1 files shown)


kimroen:
I think in general that the `--json` option should be available on most–if not all–commands. For commands that return a value it can return an object directly, and for other commands (like `ember serve`) it could return a JSON Stream. ([Mocha has a JSON Stream reporter, for an example](http://mochajs.org/#json-stream-reporter)). I realize that the way addons provide output to the user is mostly ad-hoc so this could be tricky to address, but it's so much better to address it now rather than later.

stefanpenner:
I think it's important to error if --json isn't available. Or tooling may have issues

stefanpenner:
I'm all for defining an internal structure. Then multiple printers. Json printer, ansi color printer, text printer

kellyselden:
I've started putting something together for this. You can see my progress [here](https://github.com/kellyselden/ember-cli/tree/json-help-output).

stefanpenner:
@kellyselden rad.

kellyselden:
@thomassnielsen added in #4576

PR #13, opened 2015-05-18
Dependency management for ember-cli addons

jonathanKingston:
[Rendered](https://github.com/jonathanKingston/rfcs-1/blob/uninstall-addon/active/0000-removeAddonFromProject.md)

Relates to:
https://github.com/switchfly/ember-cli-mocha/pull/47

### DIFF
--- a/active/0000-removeAddonFromProject.md
+++ b/active/0000-removeAddonFromProject.md
@@ -1,0 +1,93 @@
+- Start Date: 2015-05-18
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+Allow addons to clean up after their removal and to remove up other conflicting packages on install.
+
+# Motivation
+
+Managing addons becomes much simpler if the author doesn't need to understand dependencies that the author has had to account for.
+This has been an issue with [ember-cli-mocha](https://github.com/switchfly/ember-cli-mocha) where qunit has conflicted.
+
+# Detailed design
+
+In an addons blueprint the author could write a `afterUninstall` hook to remove all it's dependencies:
+```
+  afterInstall: function() {
+    var addonContext = this;
+
+    return this.addBowerPackagesToProject([
+      {
+        name: 'ember-mocha',
+        target: '~0.6.2'
+      },
+      {
+        name: 'ember-cli/ember-cli-test-loader',
+        target: '0.1.3'
+      },
+      {
+        name: 'ember-cli/ember-cli-shims',
+        target: '0.0.3'
+      }
+    ]).then(function() {
+      if ('removeAddonFromProject' in addonContext) {
+        return addonContext.removeAddonFromProject('ember-cli-qunit');
+      }
+    });
+  },
+  afterUninstall: function () {
+     this.removeBowerPackagesFromProject([
+       {name: 'ember-mocha'},
+       {name: 'ember-cli/ember-cli-test-loader'},
+       {name: 'ember-cli/ember-cli-shims'}
+     ]);
+  }
+```
+
+Adding:
+- removeBowerPackagesFromProject
+- removeBowerPackageFromProject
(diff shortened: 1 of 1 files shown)


active/0000-removeAddonFromProject.md
```diff
@@ -0,0 +1,93 @@
...
+        target: '0.0.3'
+      }
+    ]).then(function() {
+      if ('removeAddonFromProject' in addonContext) {
+        return addonContext.removeAddonFromProject('ember-cli-qunit');
+      }
+    });
+  },
+  afterUninstall: function () {
+     this.removeBowerPackagesFromProject([
```
stefanpenner, line 41: I think unfortunately, we don't have enough information to know if these dependencies are in-fact not being used by the primary app.

I believe the solution will be to rely on peerDependencies in v3 of npm, provide usefull health checks to addons, and warn to the user on uninstall that they must verify and potentially remove these deps.

jonathanKingston, line 41: I think the `registerDiagnostics` of Ember doctor can solve this, unless the **Alternatives** section below is useful as a shorthand for the diagnostics perhaps?

stefanpenner, line 41: > I think the registerDiagnostics of Ember doctor can solve this

:+1: 

> unless the Alternatives section below is useful as a shorthand for the diagnostics perhaps?

tell me more

jonathanKingston, line 41: Well for example if I can have some way to express what my dependencies are as a list I think this might be much easier to manage for installing, warning about conflicts and potentially allowing later to prune the dependencies if the user wants.

Behind the scenes this could just register diagnostic checks without having to express all the syntax etc.

So by adding the following on install it could add the user added packages and then warn when using ember doctor if the user has qunit installed, it could hook into the same API behind the scenes just making it simpler to implement.

```
  dependencies: function () {
    return {
      required: {
        bower: [
          {
            name: 'ember-mocha',
            target: '~0.6.2'
          },
          {
            name: 'ember-cli/ember-cli-test-loader',
            target: '0.1.3'
          },
          {
            name: 'ember-cli/ember-cli-shims',
            target: '0.0.3'
          }
        ]
      },
      conflicts: {
        addons: [
          'ember-cli-qunit'
        ]
      }
    }
  }
```

I think this might also solve the issue of addon dependencies a little neater than peerDependencies (Which is going in 3.0.0 if I remember correctly) by potentially allowing addons not to have to install into the package.json.

jonathanKingston:
@stefanpenner I think we should probably let npm 3.0 play out and see if this is still needed after that matures. Is that ok?

stefanpenner:
@jonathanKingston ya

I believe blueprint operations may allow for this nicely https://github.com/ember-cli/rfcs/pull/8#issuecomment-168814217

A high level API, could trigger `[ember:uninstall <some-package>]` operations, and the blueprints would figure out how to order-re-order etc.

This would also allow someone theoretically to create an operation, that has custom code to make more then just declarative transformations.

In addition, I believe I would like to do the following:

flesh out `ember install` to do handle all installations many stages of this, but one highlight.
- should auto-install peerDependencies of ember-cli add-ons.
- eventually do constraint based (bundler esq) installation of ember-addon peerDependencies (maybe more aggressive?)

What does this mean? 

If I run `ember install ember-cli-qunit`, and `ember-cli-qunit` peerDepends on `qunit` and `ember-qunit-notifications`, they should be auto-installed, and auto-discovered by ember-cli.

In this world, `bower` is assumed to be gone.

stefanpenner:
based my last: https://github.com/ember-cli/rfcs/pull/13#issuecomment-168817887 i think we should explore a unified `install/uninstall` npm RFC. As we approach the end of bower usage, this will become more tenable.

stefanpenner:
Is there an interested champion for this RFC, or we will close it for now.

rwjblue:

Thank you!

PR #14, opened 2015-05-18
Packager RFC

chadhietala:
[Rendered HTML](https://github.com/chadhietala/rfcs/blob/packager/active/0002-packager.md)

### DIFF
--- a/active/0002-packager.md
+++ b/active/0002-packager.md
@@ -38,12 +38,12 @@ The Builder is a re-thinking of the existing `EmberApp` constructor function tha
 - Merges an addon's app directory with the consuming app
 - Transpiles ES2015-modules code to AMD
 
-As part of the ES2015-modules transpilation we drop a `dep-graph.json` at the root of each tree. This is used by the Pre-Packager to resolve the graph. The result of this phase is an array of Broccoli trees containing all of the built assets in the app and addons.
+As part of the ES2015-modules transpilation we place a `dep-graph.json` at the root of each tree. This is used by the Pre-Packager to satisfy the graph. The result of this phase is an array of Broccoli trees containing all of the built assets in the app and addons.
 
-At this point nothing is really different then what happens today besides dropping the `dep-graph.json` and only running the addon hooks that need to happen at the beginning of a build.
+At this point nothing is really different then what happens today besides placing the `dep-graph.json` and only running the addon hooks that need to happen at the beginning of a build.
 
 ## Pre-Packager
-The Pre-Packager is a graph resolver algorithm for javascript dependencies. Another way of think about the Pre-Packager is a complex [Broccoli-Filter](https://github.com/broccolijs/broccoli-filter) and [Broccoli-Merge-Trees](https://github.com/broccolijs/broccoli-merge-trees).
+The Pre-Packager is a graph resolver algorithm for javascript dependencies. Another way of think about the Pre-Packager is a special purpose combination of [Broccoli-Filter](https://github.com/broccolijs/broccoli-filter) and [Broccoli-Merge-Trees](https://github.com/broccolijs/broccoli-merge-trees).
 
 The Pre-Packager can resolve the following types out of the box:
 
@@ -88,7 +88,7 @@ The dep-graph.json looks like the following.
       "ember-moment/helpers/duration",
       "ember"
     ],
-    "exports: [
+    "exports": [
       "default"
     ]
   },
--- a/active/0002-packager.md
+++ b/active/0002-packager.md
@@ -4,11 +4,11 @@
 
 # Summary
 
-When packaging Ember applications utilize the dependency graph to include only portions of the dep tree actually utilized. Additionally, hooks will be exposed to enable further innovation. HTTP2 packaging, custom bundles, multiple files etc.
+When packaging, Ember applications utilize the dependency graph to include only the portions of the depedency tree actually utilized. Additionally, hooks will be exposed to enable further innovation. HTTP2 packaging, custom bundles, multiple files etc.
 
 # Motivation
 
-The current app assembling strategy is naive as it merely merges and concats inputs trees in a nearly undocumented manner. Although this simplicity has served us well, it does limit flexibility of the build process. We believe we can further factor the build process in a way the improves the current À la carte strategy, while providing additional flexibitilty when desired.
+The current application assembling strategy is naive as it merely merges and concatenates input trees in a nearly undocumented manner. Although this simplicity has served us well, it does limit flexibility of the build process. We believe we can further factor the build process in a way the improves the current à la carte strategy, while providing additional flexibitilty when desired.
 
 # Scope
 
@@ -19,11 +19,11 @@ The current app assembling strategy is naive as it merely merges and concats inp
 - Traverse a dependency graph to gain selective builds
 - Lay the infrastructure for the build-side of engines
 - deprecate the hard dependency on Bower in favor NPM
-- deprecate app.import API
+- deprecate the app.import API
 
 # Definitions
 
-- __Builder__ - Responsible for generating an array of broccoli trees that contain the app and all the addons.
+- __Builder__ - Responsible for generating an array of Broccoli trees that contain the app and all the addons.
 - __Pre-Packager__ - Responsible for resolving and reducing the graph to a tree containing only reachable modules.
 - __Packager__ - Responsible for taking the resolved tree and applying – default or user provided – concatenation strategies to the tree.
 - __Entry__ - A entry node into the graph in which you traverse down from. This can be an app or engine.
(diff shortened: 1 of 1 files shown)


active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Builder
+The Builder is a re-thinking of the existing `EmberApp` constructor function that is used in the Brocfile today.  Unlike the existing `EmberApp` constructor, the `Builder` constructor only does the following:
+
+- Discover the addons
+- Runs the addons hooks that are pre-resolved tree
+- Merges an addon's app directory with the consuming app
+- Transpiles ES2015-modules code to AMD
+
+As part of the ES2015-modules transpilation we drop a `dep-graph.json` at the root of each tree. This is used by the Pre-Packager to resolve the graph. The result of this phase is an array of Broccoli trees containing all of the built assets in the app and addons.
```
stefanpenner: `drop` -> `place`

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+
+At this point nothing is really different then what happens today besides dropping the `dep-graph.json` and only running the addon hooks that need to happen at the beginning of a build.
```
stefanpenner: `placing`

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+## Pre-Packager
+The Pre-Packager is a graph resolver algorithm for javascript dependencies. Another way of think about the Pre-Packager is a complex [Broccoli-Filter](https://github.com/broccolijs/broccoli-filter) and [Broccoli-Merge-Trees](https://github.com/broccolijs/broccoli-merge-trees).
+
+The Pre-Packager can resolve the following types out of the box:
+
+- `ember-addons`
+- Legacy modules from the node_modules directory (CJS)
+- ES2015 modules from the node_modules directory
+
+While the Pre-packager comes with a sane set of defaults, custom dependency resolvers can be written to allow resolution of modules outside of node_modules.
```
stefanpenner: - how are they passed to the packager ?
- what is a proposed API/example

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+
+### Dependency Resolvers
+
+There are 2 types of dependency resolvers in the pre-packager.
+
+1. Resolvers that start traversal from a static graph
+2. Resolvers that must wait till the static graph has materialized
```
stefanpenner: this section would likely be clearer if we had a name for the output of 1. as that is what 2. operators on.

Maybe a name isn't possible, but a definition is possible:

The result of 1. is an entirely reachable graph of app, engine and addon code. This graph includes all engine/app branches, but only reachable add-on branches. If this graph has missing entries – is unsatistied – we proceed to attempt to satisfy the missing dependencies from foreign sources, such as CJS or npm provided ES2015.

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+# Definitions
+
+- __Builder__ - Responsible for generating an array of broccoli trees that contain the app and all the addons.
+- __Pre-Packager__ - Responsible for resolving and reducing the graph to a tree containing only reachable modules.
+- __Packager__ - Responsible for taking the resolved tree and applying – default or user provided – concatenation strategies to the tree.
+- __Entry__ - A entry node into the graph in which you traverse down from. This can be an app or engine.
+
+# Detailed Design
+
+## Builder
```
stefanpenner, line 33: We should also mention that the builder will have the ability to consume both un-built and pre-built dependencies. This allows users to consume local branches of un-built add-ons like ember.js and only then incur the initial build penalty.

During the release process of add-ons can pre-compute:
- `dep-graph.json`
- named module-per-file AMD modules

These artifacts should likely reside in a simple & conventional location.

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+The Builder is a re-thinking of the existing `EmberApp` constructor function that is used in the Brocfile today.  Unlike the existing `EmberApp` constructor, the `Builder` constructor only does the following:
+
+- Discover the addons
+- Runs the addons hooks that are pre-resolved tree
+- Merges an addon's app directory with the consuming app
+- Transpiles ES2015-modules code to AMD
```
stefanpenner, line 39: should we provide a brief list of what it no longer does?

active/0002-packager.md
stefanpenner: `resovle` -> `satisfy`

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+
+As part of the ES2015-modules transpilation we drop a `dep-graph.json` at the root of each tree. This is used by the Pre-Packager to resolve the graph. The result of this phase is an array of Broccoli trees containing all of the built assets in the app and addons.
+
+At this point nothing is really different then what happens today besides dropping the `dep-graph.json` and only running the addon hooks that need to happen at the beginning of a build.
+
+## Pre-Packager
+The Pre-Packager is a graph resolver algorithm for javascript dependencies. Another way of think about the Pre-Packager is a complex [Broccoli-Filter](https://github.com/broccolijs/broccoli-filter) and [Broccoli-Merge-Trees](https://github.com/broccolijs/broccoli-merge-trees).
```
stefanpenner: `complex` -> `special purpose combination of`

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+  },
+  "example-app/initializers/ember-moment.js": {
+    "imports": [
+      "exports",
+      "ember-moment/helpers/moment",
+      "ember-moment/helpers/ago",
+      "ember-moment/helpers/duration",
+      "ember"
+    ],
+    "exports: [
```
stefanpenner: oops, I forgot a `"`

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+As part of the options you can set the `composeOutput` property to a function that will receive the tree and the resolved dependency graph. It's then up to you to preform the concatenation and return the the tree.
+
+### Post Build Addon Hooks
+Once the tree has been concated, we then send that tree through the addon post-build hooks. This completes the the addon hook lifecycle.
+
+# Migration Plan
+
+## Ember CLI Developers
+
+We need to move all of the client dependencies that Ember CLI relies on to start publishing to npm as addons and are resolve their dependencies within node_modules as oppossed to bower_components. We will also need to rename some of the modules as they are at unreachable code paths e.g. `ember/resolver` does not actually resolve to the ember namespace.
```
stefanpenner: we need a bower compatibility layer, because people are using bower for many things not controlled by ember-cli. That being said, `app.import` will likely just "make it work" for now.

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Apps
+
+To ease the transition away from Bower we must provide an Ember CLI addon that adds a dependency resolver for bower components.  This would use the bower.json to whitelist modules and simply pull them into the tree from bower_components. We should warn whenever a dependency is resolved this way.
```
stefanpenner, line 197: do we need a resolver, or some sort of whitelist like the old

`app.import(foo, { exports: [ .... ] })` ?

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+
+Migrating to the Packager is as simple as changing the following in your Brocfile.
+
+```
+var Packager = require('ember-cli-packager');
+var packager = new Packager({
+  // Any exisiting EmberApp configuration plus the follow new options
+  entries: ['my-app'], // Optional if you're only building your app
+  strategies: ['default'], // Optional unless you want to change the concating
+  composeOutput: function(tree, graph) {} // Optional
```
stefanpenner, line 207: `compose` vs `composeOutput` ?

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+});
+```
+
+Since we are traversing a graph `app.import` for javascript dependencies should be removed.
```
stefanpenner: likely gotta keep these around for legacy files in vendor and bower_components. Unless you have a better idea? 

Maybe we explore some simple shimming API/experience ?

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Addons
+
+During this process we have slightly tweaked the structure of the app and addon folders in an addon.
```
stefanpenner, line 215: cc @rwjblue

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+
+### App Directory
+For the app `pods` is now a directory and will get merged into the consumers namespace under `app/pods`.
```
stefanpenner, line 218: we likely need to think about this more.

active/0002-packager.md
```diff
@@ -0,0 +1,257 @@
...
+
+### Addon Directory
+Instead of only having 1 namespace per addon we have changed the directory structure to allow for N namespaces from an addon. 
+
+The new directory structure looks like the following:
+
+```
+my-addon/
+  addon/
+    my-namespaced-thing/
```
stefanpenner, line 228: also, `my-namespaced-thing.js` would be the "bucket" where the `foo` from  `import foo from 'my-namespaced-thing'` comes from.

dschmidt:
1. Is this only about `import`ed dependencies or are you already going to handle dependencies of templates in this step (like helpers/components/etc)? Tbh I don't know how HTMLbars templates work: Do they already import everything that is needed so it's not even a special case for you to handle?

```
App and addon resolution uses the dep-graph.json to start the traversal. By default, the Pre-Packager uses the App as the entry into the graph, but developers can supply N entries. More on this later, but for now you should think of an entry as a large functional area of an application (the admin section, the main app, etc). 
```

and

```
  entries: ['my-app'], // Optional if you're only building your app
```

Can you elaborate a bit more on this? Are you planning on enabling mutliple apps per project? I want to create embeddable widgets off a large application. It of course should not load all the templates and dependencies of the main application.
I'm currently checking out options for this, engines might be one solution when they land - but that seems a bit farther ahead than this.
Moving everything that is shared between main app and widgets into an addon is very cumbersome and seems hackish if I want to keep a certain level of convernience while developing both apps at once.
So if this would enable me to build an `app.js` and a `widgets.js` with corresponding `.html` files in the same project but based on different entry points, that would be more than awesome.

rwjblue:

Thank you!

PR #16, opened 2015-05-26
Create 0000-daemonize.md

buschtoens:
[RFC](https://github.com/buschtoens/rfcs/blob/daemonize/active/0000-daemonize.md)

### DIFF
--- a/active/0000-daemonize.md
+++ b/active/0000-daemonize.md
@@ -1,0 +1,76 @@
+- Start Date: 2015-05-26
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+ember-cli will run as a daemon in the background, rather than being started for each `ember` command.
+
+# Motivation
+
+Currently running an `ember` command can take quite some time, especially when a lot of addons are installed.
+This is mainly due to the "boot up" period, when Node.js / io.js loads ember-cli itself and *all* the addons.
+Node.js / io.js unnecessarily compiles / loads a huge amount of files that very likely aren't relevant to the task.
+
+# Detailed design
+
+Instead of rebooting ember-cli everytime an `ember` command is executed, there would be a service daemon running in the background
+that performs the tasks. The major advantage is that his daemon would only have to load *once* and could run all subsequent
+tasks in an instant.
+
+The `ember` binary is merely interface to the daemon. It's sole task is proxying the entered task to the daemon,
+which then in turn does all the heavy lifting.
+
+# Drawbacks
+
+##### Yet another background service
+
+However, this isn't a backgroudn service that auto starts on system boot, but only when it is invoked by the user.
+
+##### Bigger risk of memory leaks
+
+Memory leaks would add up over time.
+
+##### Different UX
+
+Users would have to learn how to and get accustomed to start and stop the daemon, depending on how ths is implemented.
+
+However, users generally run `ember serve` and keep it running in a different terminal in the background.
+They should be used to having `ember` run in the background.
+
+# Alternatives
+
+##### Change the way addons are loaded
+
+The addon loader and addon blueprint could be changed in a way that would allow addons to export their commands and blueprints,
+without having to load the whole addon. AFAIK an addon is currently fully loaded even if a) it doesn't need most of
+the functionailty for the given task (`ember delete`) or b) it isn't even relevant for the given task (like `ember version`).
+
+The major drawback would be that all addons would have to be rewritten and that there still is a boot up period, although it'd be shorter.
+
+# Unresolved questions
+
+##### How to actually implement this?
+
+Sadly I don't know much about the technical intricacies of ember-cli, but I could imagine, that for the surface API for addons
+not much (if not nothing) would have to change.
+
+##### How would the background service be invoked by the user? How and when would it be shut down?
+
+I see two possibilities here.
+
+- Automatically start up (if not already running), whenever an `ember` command is invoked.
+  Shut down automatically after x minutes of user inactivity.
+  Also shutdown, when the user runs `ember shutdown`, or something similar.
+  
+- Start up with the command `ember daemon` (or similar) and *don't fork to the background but keep running in the foreground*.
+  This would be very much like `ember serve`. The user would close the daemon with `Ctrl + C`.
+  The user could also easily see the log output, by just switching to the terminal.
+
+##### How would the cli connect to the background service?
+
+On Unix systems that could be easily done with Unix sockets. The daemon would create a hidden `.ember-cli.sock` file
+in the root directory of the project (name irrelevant). This file serves as a [Unix socket](https://nodejs.org/api/net.html#net_net_createserver_options_connectionlistener).
+
+However, there is no such thing as Unix sockets on Windows machines. In this case, I would recommend starting the server
+on a random port and writing that port number to `.ember-cli.port` (name irrelevant).


active/0000-daemonize.md
```diff
@@ -0,0 +1,76 @@
...
+tasks in an instant.
+
+The `ember` binary is merely interface to the daemon. It's sole task is proxying the entered task to the daemon,
+which then in turn does all the heavy lifting.
+
+# Drawbacks
+
+##### Yet another background service
+
+However, this isn't a backgroudn service that auto starts on system boot, but only when it is invoked by the user.
```
stefanpenner, line 28: typoe `backgroudn` -> `background`

active/0000-daemonize.md
```diff
@@ -0,0 +1,76 @@
...
+# Detailed design
+
+Instead of rebooting ember-cli everytime an `ember` command is executed, there would be a service daemon running in the background
+that performs the tasks. The major advantage is that his daemon would only have to load *once* and could run all subsequent
+tasks in an instant.
+
+The `ember` binary is merely interface to the daemon. It's sole task is proxying the entered task to the daemon,
+which then in turn does all the heavy lifting.
+
+# Drawbacks
```
stefanpenner, line 24: another drawback is node + cross platform process management is janky.

active/0000-daemonize.md
```diff
@@ -0,0 +1,76 @@
...
+
+However, this isn't a backgroudn service that auto starts on system boot, but only when it is invoked by the user.
+
+##### Bigger risk of memory leaks
+
+Memory leaks would add up over time.
+
+##### Different UX
+
+Users would have to learn how to and get accustomed to start and stop the daemon, depending on how ths is implemented.
```
stefanpenner, line 36: I think they only need to get used to `stopping` it, as it should auto-start on issue of a command that requires it like `build` or `start`

active/0000-daemonize.md
```diff
@@ -0,0 +1,76 @@
...
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+ember-cli will run as a daemon in the background, rather than being started for each `ember` command.
+
+# Motivation
+
+Currently running an `ember` command can take quite some time, especially when a lot of addons are installed.
+This is mainly due to the "boot up" period, when Node.js / io.js loads ember-cli itself and *all* the addons.
```
stefanpenner, line 12: I don't actually think this is correct as `ember-cli` legitimately starts in a few hundred milliseconds. All the commands that take time, are because they are doing something expensive, like building/rebuilding.

If performance on those commands is an issue (which I believe it is) they should be improved directly.

My work on zeus (the fast-boot for rails) https://github.com/burke/zeus, was to deal with the extremely slow `require` times as `$LOAD_PATH` grew. This was largely due to some extremely ruby/rails design choices. Currently ember-cli does not demonstrate this issue at a magnitude that I believe actually causes grief. (Atleast as far as I am aware of)

I have been quite careful to ensure, various costly dependencies like `npm` are lazily required see: https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/npm-task.js#L18

We also try to tune this on a regular basis by using https://github.com/ember-cli/ember-cli/blob/master/package.json#L148

active/0000-daemonize.md
```diff
@@ -0,0 +1,76 @@
...
+Node.js / io.js unnecessarily compiles / loads a huge amount of files that very likely aren't relevant to the task.
+
+# Detailed design
+
+Instead of rebooting ember-cli everytime an `ember` command is executed, there would be a service daemon running in the background
```
stefanpenner, line 17: This isn't really detailed design. Rather some high level ideas. This will need to be much more thoroughly flushed out

active/0000-daemonize.md
```diff
@@ -0,0 +1,76 @@
...
+
+# Summary
+
+ember-cli will run as a daemon in the background, rather than being started for each `ember` command.
+
+# Motivation
+
+Currently running an `ember` command can take quite some time, especially when a lot of addons are installed.
+This is mainly due to the "boot up" period, when Node.js / io.js loads ember-cli itself and *all* the addons.
+Node.js / io.js unnecessarily compiles / loads a huge amount of files that very likely aren't relevant to the task.
```
stefanpenner, line 13: Is this actually the case? I have yet to see something seriously slow here, that would warrant the extra complexity.

stefanpenner:
It seems like we should focus on actually addressing any performance issues. Rather then jumping to a complicated multi-process design.

jakecraige:
I agree with Stef in that focusing on any performance issues is likely the best approach to try first. 

This would add a lot of overhead and likely a lot of confusion for new users of ember-cli and ember so ideally we want to avoid that as much as possible.

active/0000-daemonize.md
buschtoens, line 13: At least this is what I experienced. Running an `ember` command gets slower the more addons I have installed.

However, this could just be me, as I'm running ember-cli in a Ubuntu VM, which may degrade performance to a point, where this effect is noticeable.

buschtoens:
This RFC is mainly just a more elaborate issue, as I felt that this was more suitable form to present this idea.

active/0000-daemonize.md
buschtoens, line 17: Yup, you're right. This is primarily a more elaborate issue and really just a request for comments on this general idea, as I felt like it could solve those problems. :smile:

stefanpenner:
> This RFC is mainly just a more elaborate issue, as I felt that this was more suitable form to present this idea.

You should likely include detailed examples that pin-point the actual performance problems.

I have an app:
- 85,000 app loc
- 48,201 test loc
- 46 ember addons
- 51 bower addons

```
time ember -v
> ember -v  0.52s user 0.11s system 104% cpu 0.675 total
```

```
± % time ember g component foo-bar                                                                                                                                            !12833
version: 0.2.3

A new version of ember-cli is available (0.2.5). To install it, type ember update.
installing
  create app/components/foo-bar.js
  create app/templates/components/foo-bar.hbs
installing
  create tests/unit/components/foo-bar-test.js
ember g component foo-bar  0.55s user 0.10s system 103% cpu 0.608 total
```

these numbers appear reasonable, but I believe with some careful refactoring we can reduce it further. That is without introducing a complex multi-process setup.

Can you share your pain points and hard numbers?

buschtoens:
**Fresh app**

```
$ time ember generate route test-route
version: 0.2.5
installing
  create app/routes/test-route.js
  create app/templates/test-route.hbs
installing
  create tests/unit/routes/test-route-test.js

real    0m1.976s
user    0m1.868s
sys     0m0.248s
```

**Fresh app**, but with addons

```
$ time ember generate route test-route
version: 0.2.5
installing
  create app/test-route/route.js
  create app/test-route/template.hbs
installing
  create tests/unit/test-route/route-test.js

real    0m3.598s
user    0m3.464s
sys     0m0.400s
```

I find the difference noticeable, but it isn't as extreme as it was on my old development laptop. On my old laptop it sometimes took up to ten seconds for the task to execute. I don't have it here right now, but I could run a test later, if wanted.

stefanpenner:
> I find the difference noticeable, but it isn't as extreme as it was on my old development laptop. On my old laptop it sometimes took up to ten seconds for the task to execute. I don't have it here right now, but I could run a test later, if wanted.

Can you investigate what is actually causing the slow down? Which files / libs / function calls are causing the issue etc?

As you can see, with my above massive app, I do not see these same numbers. I would like to understand what the root cause is.

stefanpenner:
the largest offenders appear to be: 

```
20 '/Users/stefanpenner/src/ember-cli/node_modules/lodash/collection/find.js'¬
25 '/Users/stefanpenner/src/ember-cli/node_modules/configstore/node_modules/js-yaml/lib/js-yaml/loader.js'¬
27 '/Users/stefanpenner/src/ember-cli/lib/models/command.js'¬
27 '/Users/stefanpenner/src/ember-cli/node_modules/configstore/node_modules/js-yaml/index.js'¬
27 '/Users/stefanpenner/src/ember-cli/node_modules/configstore/node_modules/js-yaml/lib/js-yaml.js'¬
29 '/Users/stefanpenner/src/ember-cli/node_modules/markdown-it-terminal/lib/markdown-it-terminal.js'¬
35 '/Users/stefanpenner/src/ember-cli/node_modules/configstore/index.js'¬
39 '/Users/stefanpenner/src/ember-cli/node_modules/markdown-it/lib/index.js'¬
40 '/Users/stefanpenner/src/ember-cli/node_modules/markdown-it/index.js'¬
43 '/Users/stefanpenner/src/ember-cli/node_modules/markdown-it-terminal/index.js'¬
62 '/Users/stefanpenner/src/ember-cli/node_modules/leek/node_modules/request/request.js'¬
73 '/Users/stefanpenner/src/ember-cli/node_modules/leek/node_modules/request/index.js'¬
85 '/Users/stefanpenner/src/ember-cli/lib/models/project.js'¬
86 '/Users/stefanpenner/src/ember-cli/lib/utilities/markdown-color.js'¬
126 '/Users/stefanpenner/src/ember-cli/lib/models/blueprint.js'¬
127 '/Users/stefanpenner/src/ember-cli/lib/commands/destroy.js'¬
285 '/Users/stefanpenner/src/ember-cli/lib/cli/index.js'¬
```

I suspect, an easy win would be to make commands loaded on-demand rather then eagerly.
Please note, these numbers are commutative cost of loading a module, so if X requires Y and Y costs 10, X will cost cost(X) + cost(y).
- [ ] make commands load lazily
- [ ] explore the costs of markdown-color
- [ ] do the analytics also cause grief?

buschtoens:
> Can you investigate what is actually causing the slow down? Which files / libs / function calls are causing the issue etc?

How would you recommend going about that?

> I suspect, an easy win would be to make commands loaded on-demand rather then eagerly.

I would very much like that! But that would force addon developers to rewrite their plugins, wouldn't it?

stefanpenner:
> Can you investigate what is actually causing the slow down? Which files / libs / function calls are causing the issue etc?
> How would you recommend going about that?
> 
> I suspect, an easy win would be to make commands loaded on-demand rather then eagerly.
> I would very much like that! But that would force addon developers to rewrite their plugins, wouldn't it?

these are two things that should be identified before writing an RFC.

Typical flow for an RFC:
1. identify a problem
2. discover root cause
3. propose improvement via RFC

Without this, speculative solution for an RFC provides very little value and is mostly distracting and time consuming for myself and other reviewers.

buschtoens:
Closing for obvious reasons, but maybe this will be interesting in the future. :)

Anyway, how would you recommend profiling the performance of `ember-cli`?

PR #17, opened 2015-06-05
Move and Rename commands

trabus:
[rendered](https://github.com/trabus/rfcs/blob/move-rename/active/0000-move-rename.md)

### DIFF
--- a/active/0000-move-rename.md
+++ b/active/0000-move-rename.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-Provide both an `ember move` and an `ember rename` command for ember-cli, providing the ability to move/rename a file as well as update any references to it in a project. The move command will be lower level, with the rename command being more user friendly. The commands will intially be developed as an addon and can later be brought into ember-cli.
+Provide both an `ember move` and an `ember rename` command for ember-cli, providing the ability to move/rename a file as well as update any references to it in a project. The move command will be lower level, with the rename command being more user friendly. The commands will initially be developed as an addon and can later be brought into ember-cli.
 
 # Motivation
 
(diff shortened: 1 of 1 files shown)


active/0000-move-rename.md
```diff
@@ -0,0 +1,175 @@
...
+Because the mv command has a long established history, the limited functionality of this implementation could cause confusion for many users. There also may be many corner cases that won't immediately be covered by the command, possibly breaking apps, causing more issues than it solves. As long as the command leaves a trail and log of what it did, undoing any accidental changes should be managable.
+
+If the mv command is used to update several files in sequence, it will not be very efficient, as it will be repeating the same task several times without the ability to consolidate. However, performance is less of a concern in regard to migration/refactoring, as the user shouldn't need to do so frequently.
+
+# Alternatives
+
+Ember-cli alternatives are unknown, the only way to do this currently is manually through the command line and with a text editor with search and replace capabilities.
+
+# Unresolved questions
+
```
stefanpenner, line 174: - does this rename components in `HBS` files?
- does this rename/move route information in config/router.js?

stefanpenner:
lots of good detail here, but it seems like we may need to explore in-more detail how various transformations (in-file) occur.

Also, where are transform operations defined? We also don't want each transform to re-invent all source-transforms, some interopt with route-generator etc. will likely need to exist.

trabus:
> lots of good detail here, but it seems like we may need to explore in-more detail how various transformations (in-file) occur.
> 
> Also, where are transform operations defined? We also don't want each transform to re-invent all source-transforms, some interopt with route-generator etc. will likely need to exist.

My initial work on `ember-cli-mv` stalled because of exactly those questions. I ran into some dangerous/destructive situations that I hadn't accounted for in the initial spec. I will need to revisit this soon with these questions in mind.

stefanpenner:
cc @ladyleet

ladyleet:
@stefanpenner 👍

jbandura:
hi, just wanted to ask if this is still under consideration? From my side it looks super helpful and could make developers life easier (especially renaming/moving the components tests, templates and js file).

Turbo87:
closing due to inactivity

once MU lands this will become a lot easier since we only have to move/rename folders then

PR #21, opened 2015-07-12
use devDependencies from another package.json file

noslouch:
[rendered output](https://github.com/noslouch/rfcs/blob/master/active/0000-alternate-dependencies.md)

### DIFF
--- a/active/0000-alternate-dependencies.md
+++ b/active/0000-alternate-dependencies.md
@@ -4,7 +4,11 @@
 
 # Summary
 
-Allow ember-cli to target another package.json file when checking its dependency list. Normally for node projects you can symlink your local node_modules folder to another location. If ember-cli's local package.json dep list is different from what's installed in that "shared" node_module directory, it will throw an error and not build. This would allow app authors to point their devDependencies (and other keys as well, in theory) to package.json located elsewhere.
+Allow ember-cli to target another package.json file when checking its dependency list. 
+
+Normally for node projects you can symlink your local `node_modules` folder to another location. If an ember-cli project's local package.json `devDependencies` list is different from what's installed in that other `node_modules` directory, ember-cli will throw an error and not complete the command. 
+
+This would allow app authors to tell ember-cli to look elsewhere for its `devDependencies` key (and other keys as well, in theory).
 
 # Motivation
 
@@ -28,7 +32,9 @@ complicated-web-app/
     |__package.json
     |__node_modules/
 ```
-During ember-cli's intialization (before any commands are actually run), the instantiates a new Project object, which maintains a series of properties determined at runtime. One of those properites is `this.pkg`, which is the local package.json file's keys and values.
+During ember-cli's intialization (before any commands are actually run), a new `Project` object is instantiated. This object maintains a series of properties determined at runtime. One of those properites is `this.pkg`, which is the local package.json file's keys and values.
+
+`this.pkg.devDependencies` is set during initialization and used for various checks before any commands are run.
 
 This proposal would provide for a test at initialization time (`new Project()`) to look for a special key in the local package.json filed named `alternateKeys`.
 
--- a/active/0000-alternate-dependencies.md
+++ b/active/0000-alternate-dependencies.md
@@ -1,0 +1,86 @@
+- Start Date: (2015-07-12)
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+Allow ember-cli to target another package.json file when checking its dependency list. Normally for node projects you can symlink your local node_modules folder to another location. If ember-cli's local package.json dep list is different from what's installed in that "shared" node_module directory, it will throw an error and not build. This would allow app authors to point their devDependencies (and other keys as well, in theory) to package.json located elsewhere.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


noslouch:
sample implementation at [ember-cli/ember-cli#4429](https://github.com/ember-cli/ember-cli/pull/4429)

active/0000-alternate-dependencies.md
```diff
@@ -0,0 +1,92 @@
...
+    "grunt": "0.4.5",
+    // other complicated-web-app and ember deps
+    },
+}
+
+// complicated-web-app/ember-app-one/package.json
+{
+  "name": "ember-app-one",
+  "version": 0.1.1,
+  "alternateKeys": [
```
stefanpenner, line 61: `foreignProperties` maybe? `keys` isn't really a javascript term.

Also, as it shouldn't collide, we could use `_foreignProperties` ?

stefanpenner, line 61: Im also not quite sure how this would work with `npm` it feels pretty unexpected, that running `npm install` in a directory with `package.json` wont work.

active/0000-alternate-dependencies.md
```diff
@@ -0,0 +1,92 @@
+- Start Date: (2015-07-12)
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+Allow ember-cli to target another package.json file when checking its dependency list.
```
stefanpenner, line 7: I like the idea of having alternative package.json files, might be nice for the ember-try project

cc @kategengler @mike-north

active/0000-alternate-dependencies.md
```diff
@@ -0,0 +1,92 @@
...
+$ cd ember-app-one
+$ ln -s ../node_modules node_modules
+$ cd ../ember-app-two
+$ ln -s ../node_modules node_modules
+```
+
+# Drawbacks
+
+- Adds some code to the `Project` initilization process
+- It might be a rare use-case.
```
stefanpenner, line 83: also seems strange from an `npm` standpoint, as running `npm install` or `ember install` in those locations wont behave as expected. Any ideas?

kategengler:
I agree that this raises a lot of questions about `npm`. 

I'm not sure the implications for `ember-try` since `npm` support in it doesn't fully exist yet.

I've often thought it would be nice to be able to link multiple sets of dependencies -- it would allow `ember-cli` to provide an `ember-cli-deps-package.json` and prevent developers from having to merge `package.json` when upgrading, but in my mind that's always been a feature that would belong to `npm`.

active/0000-alternate-dependencies.md
kellyselden, line 61: npm 3 prepends a lot of underscored properties to the flattened package.jsons ("_phantomChildren", "_requiredBy", etc). so maybe future collisions with that approach too.

noslouch, line 61: `_foreignEmberProperties`?

active/0000-alternate-dependencies.md
noslouch, line 83: The thing about `ember install` or `npm install` not running properly in those directories follows the logic behind this rfc: that the dependencies you need for this ember app are not installed in this directory, they're installed where `_foreignEmberProperties` points, so you should be installing dependencies there.

noslouch:
I'm not familiar with `ember-try`, but I'll look into it.

The use-case that motivates this for me is another "parent app" has a combined set of dependencies for multiple apps, and that "parent app" is where a dev would run any `install` commands. 

I'm wondering if it's reasonable to expect that someone who uses this feature would be doing so after careful thought; meaning if I'm using this feature, it will be "top of mind" from which directory I should be running `npm install` and `ember install` and such.

active/0000-alternate-dependencies.md
noslouch, line 61: also, FWIW, when I say keys, I'm thinking [`Object.keys()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys)

stefanpenner:
@noslouch so the alternative to this, is to have multiple apps all with fleshed out package.json ?

noslouch:
@stefanpenner if I'm understanding you correctly, the answer is yes. the alternative to this proposal would mean each ember app has its own fleshed out package.json with its own distinct set of `devDependencies` to be maintained.

stefanpenner:
I'm torn, i do see some value in this, but I am nervous to add extra complexity here to support an edge-case, which itself has some edge-cases.

One thought: I wonder if we could expose the right hooks/api, to enable this to be a user-land extension?

noslouch:
I think the right hooks would be fine. Is there already an established pattern for `ember-cli` hooks? How would one use such an API?

Having said that, my [sample implementation](https://github.com/ember-cli/ember-cli/pull/4429/files#diff-3677132bf6240059b3a3637716019c84R50) is already hook-ish, in that `ember-cli` will operate as normal unless a special foreign dependencies key is present.

stefanpenner:
Unfortunately i don't feel comfortable adding this complexity. The combination of package.json + inherited package.json, and the existence of how the node resolution algorithm already works would create a rather complicated experience.

If this is demonstrated as a fleshed out stable, and semi popular addon we can re-evaluate it as part of the default offering.

noslouch:
That's fine. My use case eventually resolved itself by the old tried and true "just combine your apps" method

PR #22, opened 2015-08-15
Adding in JSON CSP lifecycle

jonathanKingston:
As discussed with @stefanpenner and @rwjblue - related to https://lists.w3.org/Archives/Public/public-webappsec/2015Aug/0073.html

### DIFF
--- a/active/0000-json-csp-lifecycle.md
+++ b/active/0000-json-csp-lifecycle.md
@@ -1,0 +1,105 @@
+- Start Date: 2015-08-15
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+This request is to add in the ability for addon developers to publish their CSP requirements in a standardised format. Application authors can use this to maintain a valid CSP policy. 
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


rwjblue:
@jonathanKingston - Where are we on this?  Can we progress or is the proposal stalled?

jonathanKingston:
Yay clearing out my email was useful - just seen this one.
I started some work on this and will continue it. There isn't really anything blocking it other than my time being all over the place at the moment.
I will reach out after the holidays to give you an update of where I am on this (Should be second week in Jan).

There were some discussions that were not added that were discussed at EmberCamp like CSP.json.lock which could be the most up to date CSP.json after merging in the latest addons however this may just be ignored and only adds the ability to create warnings to the user.

sandstrom:
@jonathanKingston Together with rwjblue I've updated the ember-cli-content-security-policy addon, such that it now handles an array syntax (which coincidentally is what you're using in this RFC). Just to let you know.

stefanpenner:
Im going to close this for now, we have stepped back from bundling CSP by default. I think this should be explored, if there is a champion we can reopen.

PR #23, opened 2015-08-18
Add shell completion

lazybensch:
[RFC](https://github.com/lazybensch/rfcs/blob/add_shell_completion/active/0000-command-line-completion.md) & [PR](https://github.com/ember-cli/ember-cli/pull/4709)

### DIFF
--- a/active/0000-command-line-completion.md
+++ b/active/0000-command-line-completion.md
@@ -1,0 +1,29 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+One para explanation of the feature.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Detailed design
+
+This is the bulk of the RFC. Explain the design in enough detail for somebody familiar
+with the tool to understand, and for somebody familiar with the implementation to implement.
+This should get into specifics and corner-cases, and include examples of how the feature is used.
+
+# Drawbacks
+
+Why should we *not* do this?
+
+# Alternatives
+
+What other designs have been considered? What is the impact of not doing this?
+
+# Unresolved questions
+
+What parts of the design are still TBD?
--- a/active/0000-command-line-completion.md
+++ b/active/0000-command-line-completion.md
@@ -1,14 +1,14 @@
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- Start Date: 2015-08-18
 - RFC PR: (leave this empty)
 - ember-cli Issue: (leave this empty)
 
 # Summary
 
-Adds command line completion *(tab completion)* to ember-cli that fills in partially typed commands by the user and suggests available sub commands or options. *(from now on i will refer to "commands" as in "strings in the command line" to "cli-commands" and "commands" like "generate" and "new" as unprefixed "commands")*
+Adds command line completion *(tab completion)* to ember-cli that fills in partially typed commands by the user and suggests available sub commands or options. *(from now on i will refer to "strings in the command line" as "cli-commands" and "generate" and "new" as unprefixed "commands")*
 
 # Motivation
 
-With all the already existing commands and especially all blueprints, plus the fact that any addon can add even more blueprints to your tool kit, users can get overwhelmed. Currently, when you want to execute a specific task and you don't quite know the correct cli-command you have to invoke `ember help` which is noisy and slow *( especially when you just want to know the spelling of a specific thing)*. This feature will enable the user to chose from all existing cli-commands by pressing __[tab]__ or just to let ember-cli fill partially typed cli-commands for speed.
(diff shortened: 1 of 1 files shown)


active/0000-command-line-completion.md
```diff
@@ -0,0 +1,106 @@
...
+
+  ```
+    $ ember g resource post --<tab>
+    --dry-run         --in-repo-addon=  --verbose
+    --dummy           --pod
+  ```
+
+## generation function
+
+For a good user experience we don't want to figure out those suggestions on runtime or the completion feature would not be substantially faster then the `ember help` command. So there will be a __generation function__ that generates a JSON file once after ember-cli is installed and then during every `ember install some-addon` command to ensure that blueprints added by new addons are recognized aswell.
```
stefanpenner, line 59: can you describe the lifetime of this file? Where will it live etc.

stefanpenner:
I am very +1 on this.

There is overlap with: https://github.com/ember-cli/rfcs/pull/12 and https://github.com/ember-cli/ember-cli/pull/4576  It seems both depend on a good internal structure to share and to share quickly.

cc @kellyselden

active/0000-command-line-completion.md
stefanpenner, line 59: The cache key should be described in more detail, `ember install` is only run on the installers machine now on there co-workers machines.

It may be interesting to compute a runtime cache key that represents all add-ons and there versions?

My main concern here (as with all caching) is being confident in the invalidation strategy.

lazybensch:
_**How to cache**_

@stefanpenner `ember install` not running on every machine is a good point, haven't thought of it. Assuming that caching those commands inside a json file is the best idea, then we could run the generation function as a hook on `npm install` (without arguments) afaik npm lets you do this and that is, for every user, a necessary step to alter their available commands. _(edit: i tested this and it works really good, but only if you invoke `npm i` without arguments so it would not be a silver bullet but we could use it in combination with `ember install` or something else)_

Another idea would be to invalidate the file every time you execute an ember command. So while you typing the command, ember-cli is reading from the currently cached file and once you hit enter it spawns a process that will regenerate the file in the background while ember-cli is executing whatever you typed _(the generation currently only takes about 1100ms)_

_**Or not to cache**_

Regarding the necessity of a cache itself: as i said above, currently the generation takes around 1 second, that is long enough that we do not want to run it every time a user hits `<tab>`. **BUT** currently the generation function is way more powerful than it needs to be! It is totally agnostic of commands and blueprints. The way it works is that it has a fixed root object (the _ember_ command) that has all the available commands as sub-commands. But from here on every command is deciding on its own what its subcommands are. So for example the `generate` command has a method `cliCommands` where it specifies that its own subcommands are "all blueprints", and the `destroy` method does this again. The `help` method also specifies custom `cliCommands` in listing all `commands`, including `generate` and `destroy` which in turn, again have to specifiy their `cliCommands` and parse all blueprints (that way you have autocompletion for `ember help generate model`).

The reason i chose this approach is because its very flexible. If you create a blueprint that takes a very powerful DSL as arguments, you could write your custom `cliCommands` method for that blueprint that provides shell completion for your DSL.

**However** if we sacrifice that power and just agree that no command has sub-commands, except for generate, destroy and help. We could hardcode that and speed up the generation quite a bit, maybe making it fast enough to compute on runtime. (blueprints would be gathered and parsed 1 time instead of 4, and commands 1 time instead of 2..)

PS: personally, i would rather find a good invalidation technique for a cache instead of that hard coded approach. the current logic will be a lot easier to maintain when ember-cli evolves + i might be emotionally attached to that code :D 

PPS: Another idea on how to speed up the generation so that we could do it on runtime would be to force the developer to write a dedicated `cliCommand` object for every command or blueprint or whatever he or she wants a cli-command, those objects would be lightweight enough to load them fast. On the downside, this will add redundancy cause for every `generate command` you would also need a `generate cli-command` with only the meta data interesting for shell completion and for every `model-test` blueprint you would need a  `model-test cli-command` etc. but on the upside this makes the shell completion logic more explicit in the code and we would keep all the flexibility of the feature we have right now. _(after work i will put all of these thoughts into the RFC)_

lazybensch:
small update: i was able to speed up the generation of all those sub commands from 1200ms to 40ms.

so there is no need for caching anymore we can just calculate them on the fly

stefanpenner:
> small update: i was able to speed up the generation of all those sub commands from 1200ms to 40ms.
> 
> so there is no need for caching anymore we can just calculate them on the fly

rad!

stefanpenner:
@lazybensch @kellyselden I want to move ahead with both this and https://github.com/ember-cli/ember-cli/pull/4576

There is overlap though, especially when it comes to the schema output.
Any recommendations on moving forward?

The only thing i know we should do, is to make should the JSON schema remains stable, so outside tooling and continue to rely on it.

stefanpenner:
after chatting with @kellyselden we are going to merge #4576 ultimately we need to break the tie somewhere. The goal will be to massage the PR implementing this RFC in.

Anyways, I suspect both solutions will benefit from each other. Great work guys!

kellyselden:
@lazybensch I will go through your code and look for places you can use my stuff, and vice versa.

lazybensch:
@stefanpenner i am totally fine with this! As I'm quite new to this project i'll take my time anyway.
@kellyselden you may wanna wait until tomorrow evening. i have refactored a lot locally but don't wanna push it atm. I also already had a look at you PR and feel comfortable to adjust my stuff once yours is finalized.

kellyselden:
@lazybensch Sounds good! My PR is merged now so it is as final as it's going to be.

stefanpenner:
Im very much +1 on this, I will help get the PR to completion.

PR #24, opened 2015-08-20
Blueprints for Embedded Applications

Harris-Miller:
[RFC](https://github.com/Harris-Miller/rfcs/blob/master/active/0000-blueprint-for-embedded-applications.md)

### DIFF
--- a/active/0000-blueprint-for-embedded-applications.md
+++ b/active/0000-blueprint-for-embedded-applications.md
@@ -1,0 +1,70 @@
+- Start Date: 2015-08-19
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+Ember-CLI to support embedded application blueprint.
+Generating an adjusted base application blueprint with the intent for the application to exist within another website.
+Providing an option for using Ember outside of a full website build while still leveraging ember-cli.
+Giving Ember a larger field to play on.
+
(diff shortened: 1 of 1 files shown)


csantero:
:+1:

How would this work with asset fingerprinting? When I last tried to do this, I ran into the issue that every time I built a release of the app, I would have to manually update the script tags in the hosting site's CMS to point to the new fingerprints. This was a chore and I switched to a hacky iframe solution instead.

stefanpenner:
I actually don't believe this should be a different blueprint. It really is just another build target basically any ember app can be built to.

The approaches taken for Cordova and nw.js and electron likely make the most sense here.

It is even likely a good idea to bake in the concept of a target into cli. This would likely give common infrastructure to the above mentioned solutions.

Harris-Miller:
@stefanpenner I like the idea of targets

`ember build`
`ember build --target embedded`
`ember build --target something-else`

Instead of having a different blueprint, just build to the intended target that the application will live in (e.g., full-site, embedded, etc...)

maybe `config/environment.js` could export `function(environment, target)` for build options
or have a `config/target.js` file specifically for that

stefanpenner:
i think we will take approach -> https://github.com/ember-cli/rfcs/issues/26

Harris-Miller:
:+1:

PR #28, opened 2015-11-02
Add support for app.import(…, {outputFile: …})

stefanpenner:
[rendered](https://github.com/ember-cli/rfcs/blob/app-import-output-file/active/0000-app-import-output-file.md)

### DIFF
--- a/active/0000-app-import-output-file.md
+++ b/active/0000-app-import-output-file.md
@@ -1,0 +1,94 @@
+- Start Date: 2015-11-02
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+Allow `app.import` to specify outputFIle of a given import.
+The default `app.import` would be considered to have an outputFile of
+`assets/vendor.js`
+
+# Motivation
+
+It is common for individuals to want control over the outputFile for a given
+dependency. For example, one may want to load some asm.js code independently
+rather then via the single vendor.js blob.
+
+It is also common for developers to want to group various dependencies
+together, and then lazy-load them in the routes they are required.
+
+Although not as automatic as we would like, it does provide a rather elegant
+escape valve. Further work will likely continue to explore automation.
+
+# Detailed design
+
+`outputFile:` option, specifies the target file for the given import. If
+multiple imports share an outputFile, they will be concatenated (regardless of
+type, css/images/videos/js/txt) in the order they where imported.
+
+`outputFile:` will default to `assets/vendor.js`
+
+## Examples
+
+
+#### variation 0: the default
+
+```js
+app.import('vendor/vim.js', { outputFile: 'assets/vendor.js'});
+```
+
+#### variation 1: 1 file -> 1 outputFile
+
+```js
+app.import('vendor/vim.js', { outputFile: 'assets/vim.js'});
+```
+
+* `vendor/vim.js` becomes `assets/vim.js`
+* in prod it is:
+  * uglified (unless using the uglify options it is excluded)
+  * fingerprinted (unless it is excluded via the asset-rev options)
+
+#### variation 2, multiple files to same outputFile
+
+```js
+app.import('vendor/dependency-1.js', { outputFile: 'assets/alternate-vendor.js'});
+app.import('vendor/dependency-2.js', { outputFile: 'assets/alternate-vendor.js'});
+```
+
+* in-order of the corresponding `app.import` invocation, using sourceMap
+  concat, the files are combined into `assets/alternate-vendor.js`
+  * `vendor/dependency-1.js` + `vendor/dependency-2.js` >> `assets/alternative-vendor.js`
+
+#### variation n, multiple files to same outputFile
+
+```js
+app.import('vendor/dependency-1.js', { outputFile: 'assets/alternate-vendor.js'});
+app.import('vendor/dependency-2.js', { outputFile: 'assets/alternate-vendor.js'});
+app.import('vendor/dependency-n.js', { outputFile: 'assets/alternate-vendor.js'});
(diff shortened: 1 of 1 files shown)


rwjblue:
@stefanpenner - I'm :+1: on this.  I think additional hooks would be fine, but it seems additive and can be done after this lands (IMHO).

Gaurav0:
I'm in favor of this.

You should make clear if this applies to CSS files or not.

stefanpenner:
> You should make clear if this applies to CSS files or not.

is ordered by import order based concatenation alright ?

rwjblue:
@stefanpenner - Yes, I think import order is the best we can do for CSS.

stefanpenner:
I'll call out explicitly, that it will blindly concat ANY files in import order. video/image/html/javascript/css etc.etc.

rwjblue:
Hehe, binary file concat is both a bug and a feature depending on what you want :smiling_imp:

stefanpenner:
> Hehe, binary file concat is both a bug and a feature depending on what you want

c

stefanpenner:
> You should make clear if this applies to CSS files or not.
> is ordered by import order based concatenation alright

updated

Gaurav0:
concatenated is misspelled.

stefanpenner:
> concatenated is misspelled.

can you provide line-comments, easier to track down which is mispelled

active/0000-app-import-output-file.md
```diff
@@ -0,0 +1,94 @@
...
+It is also common for developers to want to group various dependencies
+together, and then lazy-load them in the routes they are required.
+
+Although not as automatic as we would like, it does provide a rather elegant
+escape valve. Further work will likely continue to explore automation.
+
+# Detailed design
+
+`outputFile:` option, specifies the target file for the given import. If
+multiple imports share an outputFile, they will be concattenated (regardless of
```
Gaurav0: concatenated is misspelled

searls:
I'm a huge :+1: on this. I've seen two teams have to make concessions about how they bundle & distribute assets to fit the existing model, and this would have definitely helped in at least one case.

active/0000-app-import-output-file.md
stefanpenner: @Gaurav0 thank you kindly sir, updated (and hopefully fixed)

stefanpenner:
@patocallaghan / @searls either of you interested in implementing this?

active/0000-app-import-output-file.md
```diff
@@ -0,0 +1,94 @@
...
+complementary.
+
+Dropping down to broccoli is a solution available today, but for this problem,
+it feels like a slightly too low level of abstraction.
+
+# Unresolved questions
+
+* how does this relate to `type` in `app.import(..., { type: ... })` ?
+* should additional build-steps be allowed for specific output files? (I
+  suspect maybe, but a future RFC can likely explore) What parts of the design
```
Gaurav0: Unfinished question

patocallaghan:
@stefanpenner I'd be interested in taking this on once it syncs up with other stuff I have on. What sort of timeline were you thinking?

stefanpenner:
Just whenever someone has time. As usually sooner rather then later. But afew days to a week or so would be ideal.

patocallaghan:
I won't be able to commit until late next week so if someone wants to take it on beforehand, fine by me..

cibernox:
This interesting for some work I'd like to do with web workers. However I think that the addon ecosystem should be able to use this.

By example, imagine that ember data evolves to perform ajax requests and payload normalization in a web-worker. The script that this addon would generate (p.e `ember-data-worker.js`) can:

A) Contain Ember (only parts?) + Ember-data + User-defined serializers/adapters
B) Contain only User-defined serializers/adapters but use `importScripts` to load a secondary `ember-data-worker/vendor.js`) that contains Ember + Ember Data + any other dependency. 

That makes me think that addons should have the ability to add their files to a specific tree. This RFC covers part of this scenario, but not all.

stefanpenner:
Addons(who are direct descendants of the app) can already run app.import

yaymukund:
:+1: this would be useful for quickly integrating web workers

stefanpenner:
@rwjblue can you :eyes: , want to make sure i haven't missed anything.

rwjblue:
@stefanpenner - This is good to go from my perspective.  Now we need someone to implement :smiling_imp:...

MiguelMadero:
@rwjblue @stefanpenner do you still need someone to implement this? I'm working on something similar and I have some time for it. Just wanted to avoid duplicating efforts in case someone is already tackling it. Please let me know either way.

rwjblue:
@MiguelMadero - Nope, it is included in ember-cli@2.4.2

MiguelMadero:
nice! Thanks

MiguelMadero:
Is there a reason why addons aren't supported for this? 
Seems like all of the addon code goes into addons.js and later merged into vendor.js, but there isn't an option to concatenate them. I'm thinking that we could do

```
app.import('addons/ember-power-select', {outputFile: 'vendor-deferred.js'});
```

Does that sound reasonable?

stefanpenner:
that API isn't unreasonable, but it is a departure from what we do today. I think its worth exploring though, do you mind fleshing this out more in an RFC ?

MiguelMadero:
Yes. I think it's basically an extension to this RFC. Let me draft something and we can discus it there. I might need some help around the drawbacks and the potential overlap with ember-packager.

stefanpenner:
@MiguelMadero sounds good, ultimately the packager spike didn't really have a public API. Merely internal re-construction, to enable. So the high level API is fair-game.

PR #29, opened 2015-11-11
Addon black- and whitelist for EmberApps 

dschmidt:
[Rendered](https://github.com/dschmidt/ember-cli-rfcs/blob/addon-black-and-whitelist-for-apps/active/0000-addon-black-and-whitelist-for-apps.md)

### DIFF
--- a/active/0000-addon-black-and-whitelist-for-apps.md
+++ b/active/0000-addon-black-and-whitelist-for-apps.md
@@ -1,0 +1,43 @@
+- Start Date: 2015-11-11
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+It should be possible to white- and/or blacklist addons in `EmberApp`.
+
+# Motivation
+
+If there are two (or more) `EmberApp`s, it's very likely that not all applications need all addons.
+E.g. if there is a main page application and one application for embeddable widgets, the main page might need all sorts of addons like `ember-modal-dialog` which adds completely useless bytes to widgets javascript and css files for the widgets. Other addons may add useless initializers or other things that have runtime performance penalties for no benefit.
+
+# Detailed design
+
+When EmberApp ctor gets passed a blacklist like this
+
+```javascript
+  EmberApp({
+    addonBlacklist: ['ember-modal-dialog']
+  });
+```
+
+it won't add addons whose `name` matches `ember-modal-dialog` to the list of addons for this app, just as if the addon's `isEnabled()` hook returned `false`.
+
+C.f. https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L344, this is also were I would add this check.
+
+Whitelist could work analogously.
+
+# Drawbacks
+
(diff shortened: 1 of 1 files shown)


active/0000-addon-black-and-whitelist-for-apps.md
```diff
@@ -0,0 +1,43 @@
...
+
+# Drawbacks
+
+- It adds a bit of API surface while you (possibly) don't care for the multiple app use case of ember-cli.
+
+- It kinda makes the `name` of an addon public api, so people might change it, not noticing they are breaking people's build (and people might not notice it either). Some addons have names like `Ember CLI ic-ajax` which seems awkward to use as an identifier.
+
+# Alternatives
+
+- Add a unified way to enable/disable an addon via normal config
```
stefanpenner, line 38: an alternative, is to have `app2/package.json` that exports `EmberApp2` and is required by the primary app.

downside is lots of node_modules duplication. But if the 2 apps depend on the same addon but two versions it would atleast not collide, but the UI may also not actually work :P

stefanpenner:
interesting, cc @rwjblue

active/0000-addon-black-and-whitelist-for-apps.md
dschmidt, line 38: In fact, that is already possible today - and I considered doing it that way (by just putting two complete projects into one repository).
But I decided to not go down that road, it's a nightmare to maintain two package.json files, always having to remember to bump versions in both files. I'd say two versions of the same addon is a corner case in which it might make sense to do that. Noone stops anyone from doing it, but it would be nice for me not having to :-)

Still want me to add it to the RFC text as another possible option?

trabus:
I think it might be better defined through `config/environment.js`, rather than the `ember-cli-build.js` file. That way it can be controlled at the environment level. This would be beneficial for helper libs that belong only in test or dev.

stefanpenner:
> rather than the ember-cli-build.js

I think the ember-cli-build is exactly where this should be, environment.js is for client-side configuration. Not build

dschmidt:
Yeah, I agree with @stefanpenner. `ember-cli-build.js` is where you configure build options for addons, that should also be where you disable them for builds. If at all I try to pipe build configuration into runtime config, but never the other way round.

trabus:
Ah, that's a good point. I guess what I was getting at is having some way to blacklist based on the build environment, rather than all environments, but I think I was maybe misappropriating the feature for another issue that has bothered me in the past.

I saw it as a way to have a simple config to manage addon exclusion by environment rather than needing to jump through hoops to remove addons for specific environment builds.

dschmidt:
Soooooooo... should I try to implement it or does this need any "more official approval"?

stefanpenner:
@rwjblue r?

I like the idea. I don't like passing everything to ember app, but that is currently our API. Snd somewhat orthogonal to the feature so I should likely not get hung up on that

rwjblue:
r+

stefanpenner:
@dschmidt implementation seems good next steps.

PR #33, opened 2016-01-18
add-ember-data-pod-structure

CNDW:
Feature request for improved pod structure for ember-data


stefanpenner:
@rwjblue :eyes:

stefanpenner:
Thanks for the RFC, it feels like well thought out and ultimately quite good, but I will close this RFC for now due to:
- larger pod discussion is happening here: https://github.com/emberjs/rfcs/pull/124
- that RFC actually proposed something quite similar `namespace/name/type.js` or `namespace/name.type.js` I prefer the later.

I suspect that RFC will continue this effort. Feel free to jump in and provide feedback there.

Issue #26, opened 2015-09-05
build target + type


stefanpenner:
_some thoughts, leaving them here for future travelers_

---

There is a constant request to expand the number of environment one can use. Although this is perfectly reasonable and internals should (and will support this). I have noticed another similar trend, build targets 
- cordova
- nw.js
- electron
- full page web
- embedded web etc)
- etc.
- some custom mode

Ideally, I would like to not see dev_cordova prod_cordova test_corvoda, dev_electron,prod_electron etc.etc. As this quickly becomes entirely unmanageable and the cost of adding a new target becomes quite large. It also becomes tricky when dealing with addons, that maybe only need to do something in testing of Cordova.

One thought I had, was to split the concept into `target` and `type`. 
- **target**: destined platform/mode : defaults to `web`
- **type**:  more similar to todays `environment` – in-fact it may make sense just to call it environment. : defaults to dev

``` sh
ember build --target electron --type test
ember server --target electron --type test
ember test --target electron
ember test --target electron --server
```

This could allow us to provide stable public CLI interface, addon hooks, and should help become a good foundation for future innovation.

---

stuff to explore
- how does one define a target
- how are commands made aware of the target goals
- how do addons work with this.
- how do we merge "goals" of env + target at runtime.

cc people who have been doing related work (based on emberaddons.com search): @mike-north @poteto @rwjblue @jakecraige @juanazam @theodorton @brzpegasus @rmachielse

rmachielse:
Absolutely interesting! I have solved this problem right now by having a separate directory for the app I want to build. Developers can then symlink the contents of the directory with the real app directory, so they can add files or overwrite them.

Doing this with a `target` sounds more clean. I think `type` is a very general name, why not just `target` and `environment`?

Still, this means that there has to be a way to add or remove files depending on the `target`. For example by having extra optional directories per target that will be merged into app on build.

stefanpenner:
> Doing this with a target sounds more clean. I think type is a very general name, why not just target and environment?

no reason, i am fine with either actually. My only concern was migration (unless it turns out --target is merely a layer on top, which is likely)

stefanpenner:
> Still, this means that there has to be a way to add or remove files depending on the target. For example by having extra optional directories per target that will be merged into app on build.

ya this seems easy at the addon/broccoli level (assuming we expose the correct info)

juanazam:
[...]

PR #35, opened 2016-02-12
ember-cli platform flag and platform support

runspired:
Intended to supplant #26
[rendered](https://github.com/isleofcode/rfcs/blob/ember-cli-platforms/active/0000-plaforms.md)

This was extracted from an addon proposal, and there are areas (testing, how to handle security concerns of some platforms, and how to expose the pipeline to plugins) that would evolve overtime in an addon and are less fleshed out from the perspective of being an addition to ember-cli than they ought to be.

Addon where this is [likely to be built initially](https://github.com/isleofcode/ember-cli-platforms)

### DIFF
--- a/active/0000-plaforms.md
+++ b/active/0000-plaforms.md
@@ -1,0 +1,328 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+The ember-cli platform flag treats platforms as build targets, and gives you a
+workflow for producing builds customized to a platform, as well as streamlined tooling,
+enhanced resuability, and better separation of concerns for applications that
+target more than one platform.  Platforms are roughly equivalent to web views,
+or a shell which provides a web view, such as a browser.
+
+# Motivation
+
+There is a growing trend of shipping Ember applications inside of WebView shells to
+provide installable cross-platform applications and to gain access to native features.
+
+Currently, each platform has a homegrown solution for managing build/serve/test/deploy
+processes.  We could reduce the surface area of commands these platforms expose and
+ease the building of addons specific to a platform by enabling ember-cli to understand
+the existence of platforms and how various assets relate to them.
+
+# Detailed design
+
+## Usage
+
+These are the docs for using `ember-cli-platforms` with an installed platform plugin.
+For the docs on creating a platform plugin, [skip ahead](./README.md#creating-plugins).
+
+**Create a platform.**
+```cli
+ember platform <type> [-t "<name>"]
+```
+
+Generates a new platform of `<type>` with an optional `<name>` parameter used when more than
+one instance of a platform is desired.
+
+With Cordova this might look like the following.
+
+Default setup (one project)
+```cli
+ember platform cordova
+```
+
+Advanced setup (multiple projects)
+```cli
+ember platform cordova -t "android"
+ember platform cordova -t "ios"
+```
+
+For all commands, `platform` can be abbreviated as `p`, and
+commands that mirror `ember-cli` commands preserve the `ember-cli`
+shorthands for those methods as well.
+
+**Serve a specific platform.**
+```cli
+ember s -p "<type>[-<name>]"
+```
+
+This is shorthand for:
+```cli
+ember serve --platform="<type>" [--target="<name>"] --environment="development"
+```
+
+**Build a specific platform.**
+```cli
+ember b -p "<type>[-<name>]"
+```
+
+**Test a specific platform.**
+```cli
+ember t -p "<type>[-<name>]"
+```
+
+**Deploy a specific platform.**
+```cli
+ember d <deployTarget> -p "<type>[-<name>]" 
+```
+
+This is short hand for:
+```cli
+ember deploy --deployTarget="<deployTarget>" --platform="<type>" [--target="<name>"] --environment="development"
+```
+
+
+### Platform Specific Code
+
+Platforms would leverage in-repo-addon semantics and Broccoli plugins to enable
+sprinkling of platform specific code into an app as needed with "platform addons".
+
+A platform addon is generated automatically for each platform created with `ember platform`.
+
+```
+<project>
+  /lib
+    /<type>[-<name>]
+```
+
+These "addons" function differently from normal addons, and would be prefixed with `-platform-`.
+Instead of code in the app overriding or importing code from the addon, the addon would 
+take precedence.  This let's you selectively replace or add entire modules as needed.
+
+You can import the corresponding `app` version of a module into the platform addon.
+
+```js
+import Foo from '<project-name>/routes/foo';
+
+// our platform version of routes/foo
+export default Foo.extend({
+  hello: 'World!'
+});
+```
+
+### Platform Specific Config
+
+A platform config file is generated automatically for each platform you create with `ember platform`.
+```
+<project>
+  /config
+    /platforms
+      /<type>[-<name>].js
+```
+
+This configuration will be merged with your primary configuration.
+
+## Plugins
+
+Users should be able to create platform plugins as addons which implement 
+any number of specified hooks. These plugins should carry in the same tradition
+in form, function and hook names as `ember-cli-deploy`.
+
+## Creating Plugins
+
+Platforms have several forms of plugins:
+
+1. `process hooks` which add behaviors to one or more platforms.
+2. `platform plugins` which implement support and generators for a specific platform.
+3. `deploy plugins` which implement support for deployment of a platform.
+
+Platform plugins (generally) are also process hooks.
+
+## All plugins
+
+Platform plugins are ember-cli addons with a few platform specific traits:
+
+1. they contain a package.json keyword to identify them as plugins
+2. they are named ember-cli-platform-*
+3. they return an object that implements one or more of the ember-cli-platforms pipeline hooks
+
+
+### Creating a Process Hook
+
+#### Create an addon
+
+```cli
+ember addon ember-cli-platform-<name>
+```
+
+#### Mark the addon as a plugin
+
+Identify your addon as a plugin by updating the package.json like so:
+
+```js
+// package.json
+
+"keywords": [
+  "ember-addon",
+
+  // if present, indicates this addon contains process hooks for ember-cli-platforms
+  "ember-cli-platform-plugin",
+  
+  // if present, indicates this addon contains a platform implementation for ember-cli-platforms
+  "ember-cli-platform",
+  
+  // if present, indicates this addon implements deployment hooks
+  "ember-cli-platform-deploy"
+]
+```
+
+
+#### Implement one or more pipeline hooks
+
+In order for a plugin to be useful it must implement one or more of the ember-cli-platforms
+pipeline hooks.
+
+To do this you must implement a function called `createPlatformPlugin` in your `index.js` file.
+This function must return an object that contains:
+
+1. A name property which is what your plugin will be referred to.
+2. One or more implemented pipeline hooks
+
+**Example:**
+
+```js
+module.exports = {
+  name: 'ember-cli-platform-live-reload',
+
+  createPlatformPlugin: function(options) {
+    return {
+      name: "live-reload",
+
+      didBuild: function(context) {
+        //do something amazing here once the project has been built
+      }
+    };
+  }
+
+};
+```
+
+#### Process Hooks
+
+- configure
+- setup
+- willBuild
+- build
+- didBuild
+- teardown
+
+
+
+### Creating a Deployment Plugin
+
+This is built overtop of [ember-cli-deploy](http://ember-cli.com/ember-cli-deploy/docs/v0.5.x/pipeline-hooks/),
+but integrates with your platform flow.
+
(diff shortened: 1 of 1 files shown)


active/0000-plaforms.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Usage
+
+These are the docs for using `ember-cli-platforms` with an installed platform plugin.
+For the docs on creating a platform plugin, [skip ahead](./README.md#creating-plugins).
+
+**Create a platform.**
+```cli
+ember platform <type> [-t "<name>"]
+```
```
Turbo87, line 33: I'm not a big fan of `ember platform` as a "generate something for me" command, but then again we also already have `ember addon` which is sort of similar. An "addon" however is something that actually gets generated while it's not exactly a "platform" that gets generated, but a platform specialization and configuration. I'm not a native speaker but maybe we can find a better name for the command?

Also wouldn't it be better to use `ember generate platform <name>` here since the command description actually already starts with "Generates"?

active/0000-plaforms.md
```diff
@@ -0,0 +1,328 @@
...
+
+A platform addon is generated automatically for each platform created with `ember platform`.
+
+```
+<project>
+  /lib
+    /<type>[-<name>]
+```
+
+These "addons" function differently from normal addons, and would be prefixed with `-platform-`.
```
Turbo87, line 99: why do we need the prefix? if platforms must have a corresponding config file we could just list the platform config files to determine which ones are "special".

active/0000-plaforms.md
```diff
@@ -0,0 +1,328 @@
...
+Instead of code in the app overriding or importing code from the addon, the addon would 
+take precedence.  This let's you selectively replace or add entire modules as needed.
+
+You can import the corresponding `app` version of a module into the platform addon.
+
+```js
+import Foo from '<project-name>/routes/foo';
```
Turbo87, line 106: somehow I have the feeling that the implementation of this will be quite complicated. also what if you want to share code between e.g. the two cordova platforms but want a different impl on electron?

active/0000-plaforms.md
```diff
@@ -0,0 +1,328 @@
...
+  "ember-addon",
+
+  // if present, indicates this addon contains process hooks for ember-cli-platforms
+  "ember-cli-platform-plugin",
+  
+  // if present, indicates this addon contains a platform implementation for ember-cli-platforms
+  "ember-cli-platform",
+  
+  // if present, indicates this addon implements deployment hooks
+  "ember-cli-platform-deploy"
```
Turbo87, line 176: why use three different keywords here and not just `ember-platform`? we could probably just check if the hooks are implemented by the addon or not, right?

active/0000-plaforms.md
```diff
@@ -0,0 +1,328 @@
...
+- configure
+- setup
+- willBuild
+- build
+- didBuild
+- teardown
+
+
+
+### Creating a Deployment Plugin
```
Turbo87, line 222: Should we possibly limit the scope of the proposal for now and skip the deployment process for now? Do we actually need to adjust `ember-cli` for that in any way or is this only relevant for `ember-cli-deploy`?

active/0000-plaforms.md
```diff
@@ -0,0 +1,328 @@
...
+
+If a user generates a cordova platform with the following command
+
+```cli
+ember platform cordova -t "ios"
+```
+
+Then an `npm` command script will be generated and installed that proxies use of
+`cordova-ios` into the cordova project directory ios, allowing you to easily use
+the cordova-cli.
```
Turbo87, line 280: I'm not sure I understand what should be generated here...

active/0000-plaforms.md
runspired, line 33: A good point.  A key distinction is that a platform does far more than generate `test` and `app` files, you also get a platform addon similar to an `in-repo-addon` and a platform directory (e.g. `platforms/cordova-ios`.

Another potential distinction is that it may be beneficial to have `platform` have `install` behavior as well.  If a given platform addon is not already present (say `ember-cli-platform-cordova`), then ember platform would fetch and install it.  Whether this is valuable, I'm less certain, and I'd be comfortable falling back to `generate` over increasing surface area.

active/0000-plaforms.md
runspired, line 99: They don't necessarily have to have a config file, I think relying on a potentially volatile (and easily deletable) configuration file is going to be trickier than distinguishing between in-repo addons and in-repo platforms in this way.

active/0000-plaforms.md
runspired, line 106: > somehow I have the feeling that the implementation of this will be quite complicated.

This is fairly simple to pull off with a few broccoli funnels, similar to how a Dummy app works with `addon` and `app` directories in an addon.

> also what if you want to share code between e.g. the two cordova platforms but want a different impl on electron?

Sharing in this way would be neat, but trickier to work out.  Potentially we should enable a `cordova` platform that both `cordova-ios` and `cordova-android` would inherit from.  E.g., one for platform type, and an additional specialization for name.

active/0000-plaforms.md
runspired, line 222: We should limit the scope and leave this to an addon for the moment.

While this is mostly relevant for `ember-cli-deploy`, it's important to note a lot of base code is going to get repurposed for this project from ember-cli-deploy.

active/0000-plaforms.md
runspired, line 280: Projects such as cordova come with their own cli and own commands to run, ember-cli-cordova attempted to proxy these commands through an `ember` command, but that results in a very large overhead to setting up and running the command.

This would be a simple script which allows you to run that platform's commands from your ember project directory instead of needing to change directories.  It's a convenience factor which probably should be taken out of scope and left to an addon.

active/0000-plaforms.md
runspired, line 176: We can get by with one.

Ostensibly, this is for permissions elevation.  In the original conception, `ember-cli-platform` meant an addon worked with the `ember platform` command, `ember-cli-platform-plugin` meant it would work with `build`, `serve` and `test` commands, and `ember-cli-platform-deploy` meant it would work with the `deploy` command.

Keeping this distinction probably would let us be a bit faster at booting the commands, but it's definitely not essential.

kellyselden:
I didn't see anything about supported platforms. Would ember-cli come built in with any platforms, or just the base `platform` command with no implementations. Would the current ember-cli system be refactored into a default (implicit) platform of "web-app"?

Turbo87:
> Would ember-cli come built in with any platforms, or just the base platform command with no implementations.

If I understood correctly `ember generate platform cordova --name=ios` would try to install `ember-platform-cordova` automatically and then use it to generate the necessary files

> Would the current ember-cli system be refactored into a default (implicit) platform of "web-app"?

Since the usual "web-app" has no other special tooling needed it should probably stay the default. I think it's also still the 90% use case so that should be as simple as possible, which also means not bothering dev with the "platforms" concept as long as they don't need it.

runspired:
@Turbo87 @kellyselden no platforms come built in, the idea is to expose the right hooks and flags so that a platform plugin can be individually developed for each platform that can rely on a standardized setup and process instead of needing a unique DSL and much "guts" code repetition for these things.

> Would the current ember-cli system be refactored into a default (implicit) platform of "web-app"?

Yes.  End user's (addon app creators) wont see any change in code organization, but running any command without the platform flag would be the equivalent of running it with the "web" flag.

> If I understood correctly ember generate platform cordova --name=ios would try to install ember-platform-cordova automatically and then use it to generate the necessary files

This is up for debate. I think it should be the case, but that might be over greedy and lead to an undesirable level of complexity.

Turbo87:
> This is up for debate. I think it should be the case, but that might be over greedy and lead to an undesirable level of complexity.

I think I would prefer something like this:

```
ember install ember-platform-cordova
ember generate cordova-thingy android
```

It's not much more difficult than the proposed command but leverages a lot of the existing code and concepts.

runspired:
I'm going to leave this as a comment as I haven't had time to update the RFC.

For this RFC, we will need to make sure to:
- permeate the platform through the system

We likely should enable:
- multiple platforms to be served at once

After conversations with @brzpegasus @stefanpenner @lukemelia and @felixrieseberg, it looks like this RFC has surfaced a number of common addon needs that are larger than the scope here, and potentially could become 3 RFCs.

**"middleware hooks"**
-> addon hooks for testem options
-> custom testem launchers
-> customize build, and watch including with a custom index.html for the tests

The short and dirty solution for this for the near term is to allow platform plugins to extend or replace tasks more easily.

**"install patches"**
-> ability to "patch" modify
- README.md
- package.json
- testem.json
- .travis.yml / .circle.yml
- environment.js

runspired:
@Turbo87 We would still be able to leverage the existing code, but I do agree with the sentiments of not expanding the surface area.  Possibly another way of achieving this elegantly would be.

```
ember install ember-cli-platform-cordova --name="ios"
```

Where that flag is passed into the default blueprint, and (I'm pretty sure) that capability already exists.

Turbo87:
> "install patches"

I don't think there's anything stopping you from doing this right now, there's just no API available for e.g. reading/writing markdown. or are you talking about applying actual patch/diff files?

Turbo87:
> `ember install ember-cli-platform-cordova --name="ios"`

but then you need a different command if you want to add a second cordova target... I'd try to keep it as simple as possible. `install` for installing the addon, `generate` for generating a target/platform/whatever it's called.

runspired:
@Turbo87 since install runs the default blueprint, we'd be assuming that

```
ember g platform cordova --name="android"
```

would actually be the same blueprint.

runspired:
@Turbo87 

> there's just no API available

Correct. @stefanpenner and I both feel there's value in having an official API available. I've done this in the ember-cli-toolbelts experiment, and would love to expand on that work and give it a nice API.

Turbo87:
@runspired I know that it could technically work. I'm just saying that as an end user of such an `ember-platform-cordova` addon it would be easier to grasp if the commands were separate :wink:

Turbo87:
> there's value in having an official API available.

I don't see how that is the responsibility of `ember-cli`. IMHO it should give blueprints and/or addons a hook after the files are written so that they can be processed if needed, but writing markdown or something like that seems out of scope for me. there are probably libs on NPM that could do that but I don't think they should be part of every `ember-cli` install for the few addons that actually want to modify a default readme file.

active/0000-plaforms.md
runspired, line 106: > This is fairly simple to pull off with a few broccoli funnels

Actually I retract that. It's not fairly simple, but nor is it overly complex.

I imagine it would work sort of like this.

When merging the platform with the app, we rename the platform modules to `<app-name>`, and when a conflict is encountered, rename the `app` version to `app/` from `<app-name>/`.  And the import of that from the platform would be:

```
import OriginalModule from `app/path/to/original`;
```

Initially, this results in there always being extra code (e.g. both original and platform version) shipped, but once we have tree shaking (I'd imagine not very far off from the point this would land) that wouldn't matter.

kellyselden:
> Since the usual "web-app" has no other special tooling needed it should probably stay the default. I think it's also still the 90% use case so that should be as simple as possible, which also means not bothering dev with the "platforms" concept as long as they don't need it.

@Turbo87 I wasn't suggesting people would have to run extra commands, just that ember-cli internals use the platform system to generate the "web" platform implicitly, similar to @runspired 's comment:

> Yes. End user's (addon app creators) wont see any change in code organization, but running any command without the platform flag would be the equivalent of running it with the "web" flag.

nathanhammond:
> IMHO it should give blueprints and/or addons a hook after the files are written so that they can be processed if needed, but writing markdown or something like that seems out of scope for me.

@Turbo87 @runspired See: https://github.com/ember-cli/rfcs/issues/36

We're basically discovering that, if we want to keep code out of ember-cli core, we need to carefully design extensibility patterns for each major ember-cli construct. That discovery was echoed again in today's meeting.

nathanhammond:
At the conclusion of today's team meeting we were unanimous in the opinion that this is the wrong level at which to attack this problem. The underlying problem here isn't that we need ember-cli to support a new type of behavior, but instead that modifications to ember-cli are **required** to support this type of behavior. It's our goal that something of this nature should be possible to accomplish inside of an addon without requiring the present amounts of copypasta, clever imports, and monkeypatches.

At this point we know very well that it is impossible for us to do much more than make educated guesses as to how people wish to use ember-cli, and who knows what our Angular friends will discover in their world. With that context, our work on ember-cli needs to be responsible for a few things:
- Making sure that the happy path for users remains consistent, stable, and _easy_.
- Maintain focus on the core pieces of ember-cli as opposed to trying to consume responsibility for the entire ecosystem of ember-cli addons.
- Make it possible for users to rapidly get exciting new features.

In order to achieve the last one we discussed identifying/designing what the core extensibility primitive is inside of ember-cli for each of the main components. This is similar in what was discovered with Broccoli where everything along the way simply receives a tree. It's unlikely that we do as well in creating such a beautifully simple primitive to accomplish the entire functionality, but that should be the measuring stick by which we evaluate our solution for elegance.

With that, @runspired, I motion to close this in favor of a different RFC (or multiple RFCs), yet to be written. If you agree we'll close this PR. :smile:

rwjblue:

Thank you!

PR #40, opened 2016-03-21
Deprecate overridable commands

lazybensch:
[rendered](https://github.com/lazybensch/rfcs/blob/deprecate_overridable_commands/active/0000-deprecate_overridable_commands.md)


rwjblue:
Hmm, we definitely have a few scenarios where customizing internal commands is useful. For example, ember-cli-mocha adds additional options (--grep I think?) to the test command. 

I think the way that is done in ember-cli-mocha would still work fine, but I wanted to point out at least one use case where internal commands are modified...

nathanhammond:
Hey @lazybensch! We talked about this on Thursday at the ember-cli core team meeting. As our current interface allows for extending the built-in commands it would be a breaking change to modify this behavior.

Your goal is simple: performance. We believe pretty strongly that we can achieve a more-performant version of this with far less drastic measures than simply preventing extension of the base `command`s. I'd encourage you to dive in and identify what it is that is taking a long time and come up with alternative means to get to this end.

If we can't address performance issues after investigation with just pure optimization we can start talking about a declarative way of saying, "I'm going to extend this command" via some manifest or something of that nature. Something where it's still possible, but we put guidelines around it so we are able to make more assumptions is a more-viable move-forward strategy.

If, after an ​_in depth_​ exploration of performance, we're _still_ unable to meet reasonable perf targets then we'll reopen this topic for discussion. In the mean time I'm going to close this, and I look forward to working with you on PRs to get some performance fixes into ember-cli!

PR #46, opened 2016-04-04
Improved Release Process

nathanhammond:
[Rendered](https://github.com/nathanhammond/ember-cli-rfcs/blob/improved-release-process/active/0046-improved-release-process.md)

PR #47, opened 2016-04-05
Node Support Policy

nathanhammond:
[Rendered](https://github.com/nathanhammond/ember-cli-rfcs/blob/node-support/active/0047-node-support.md)

### DIFF
--- a/active/0047-node-support.md
+++ b/active/0047-node-support.md
@@ -9,15 +9,41 @@ One of the nice things about Node.js being our server environment is that we are
 
 We've avoided adding libraries like Babel into our ember-cli applications in order to maintain debuggability and reduce magic. This leaves us yearning for the features, flexibility, and syntactic sugar of more-recent V8 releases. In order to not be disruptive we have continued to support the 0.10 branch which has tied us to an ancient V8 engine.
 
+Additional issues with old versions of Node.js not related to language features:
+* Bundling old and questionably working versions of `npm`.
+* Bundling ancient and entirely unsupported versions of libuv and v8.
+* Missing some core API and performance features such as `execSync` or `vm` on Node.js 0.10.
+
 # Proposal
 
-We track the [existing LTS releases for Node.js](https://github.com/nodejs/LTS#lts_schedule), but only provide support for the active LTS releases, or the stable versions which will become LTS releases. Once an LTS is moved into maintenance mode we will drop support for that version. Versions of Ember CLI whose support crosses one of the Node.js boundaries will be supported per the ember-cli release support policy which is currently unspecified, though described by RFC #46.
+Commits to the `HEAD` of the master branch will provide support for any Active (not Maintenance Mode) Node.js LTS and the current stable Node.js version(s). This will be enforced via CI, preventing us from landing code which won't work with versions which we support. This means that our schedule for support is tied to the [LTS release schedule for Node.js](https://github.com/nodejs/LTS#lts_schedule).
 
-For our currently existing support of 0.10 and 0.12 we should drop support 2016-10-01 and 2017-04-01, respectively, at the official end of support for those releases. This gives time for our community to migrate off of those environments.
+For our currently existing support of 0.10 and 0.12 we should drop support at the end of their Maintenance Mode–their official end-of-life from Node.js. This gives time for our community to migrate off of those environments which we presently support.
+
+## Legacy Support
+
+* v0.10: LTS Maintenance Mode ends on 2016-10-01. We drop all support on that date.
+* v0.12: LTS Maintenance Mode ends on 2017-04-01. We drop all support on that date.
+
+## Ongoing support per this RFC:
+
+These examples assume major version bumps. This is not required per Node.js policy in order to become an LTS.
+
+* v4.2: LTS Maintenance Mode begins on 2017-04-01. We drop support on that date for new commits to `master`. _We must have an Ember CLI LTS strategy specified by this date._
+* v5: Stable release, never LTS. Began support 2015-10-01. All support expires on 2016-07-01.
+* v6: Released as Stable approximately 2016-04-01, converted to LTS Active Support 2016-10-01. Begin support on 2016-04-01. Active support ends on 2018-04-01 at which point we drop support on that date for new commits to `master`.
+* v7: Stable release, never LTS. Begin support upon release 2016-10-01. All support expires on 2017-07-01.
+
+# Release Process and Support Policy
+
+The Ember CLI release process and support policy is presently `undefined`. We are currently specifying a release process in RFC #46. We must specify a support policy no later than 2017-04-01. The support policy doesn't have to be complete before then as we will continue to 100% support all Node.js supported releases until that date under this RFC.
+
+This RFC is distinct from providing guarantees that some version of Ember CLI will run securely on each Node.js LTS release. Currently we provide no guarantees of support for any previously released version of Ember CLI–which is a gap we wish to correct and are working toward.
+
 
 # Drawbacks
 
-This will continue to delay adoption of new and exciting features, such as `async` and `await`, which are progressing through stages of standardization. However, it also makes it explicitly clear when we can expect to begin supporting certain features. It will also make it more difficult to support changes which need to be cherry-picked to release branches that support older versions of Node.js but we hope to write perfect code and never need to release patches to previous versions.
+This will continue to delay adoption of new and exciting features, such as `async` and `await`, which are progressing through stages of standardization. However, it also makes it explicitly clear when we can expect to begin supporting certain features. It will also make it more difficult to support security fixes which need to be cherry-picked to older release branches that support older versions of Node.js but we hope to write perfect code and never need to release patches to previous versions. We also expect that the security patches that would require this to be infrequent.
 
 # Alternatives
 
--- a/active/0047-node-support.md
+++ b/active/0047-node-support.md
@@ -1,0 +1,28 @@
+- Start Date: 2016-04-04
+- RFC PR: #47
+
+# Summary
+
+One of the nice things about Node.js being our server environment is that we are not beholden to the vast installed base of browsers and can instead move our community by fiat. However, we've never explicitly communicated what our plans for support will be and our current pattern has grown organically. We need to define these guidelines so that larger organizations are able to plan for future migrations.
+
+# Motivation
+
+We've avoided adding libraries like Babel into our ember-cli applications in order to maintain debuggability and reduce magic. This leaves us yearning for the features, flexibility, and syntactic sugar of more-recent V8 releases. In order to not be disruptive we have continued to support the 0.10 branch which has tied us to an ancient V8 engine.
+
+# Proposal
+
+We track the [existing LTS releases for Node.js](https://github.com/nodejs/LTS#lts_schedule), but only provide support for the active LTS releases, or the stable versions which will become LTS releases. Once an LTS is moved into maintenance mode we will drop support for that version. Versions of Ember CLI whose support crosses one of the Node.js boundaries will be supported per the ember-cli release support policy which is currently unspecified, though described by RFC #46.
+
+For our currently existing support of 0.10 and 0.12 we should drop support 2016-10-01 and 2017-04-01, respectively, at the official end of support for those releases. This gives time for our community to migrate off of those environments.
+
+# Drawbacks
+
+This will continue to delay adoption of new and exciting features, such as `async` and `await`, which are progressing through stages of standardization. However, it also makes it explicitly clear when we can expect to begin supporting certain features. It will also make it more difficult to support changes which need to be cherry-picked to release branches that support older versions of Node.js but we hope to write perfect code and never need to release patches to previous versions.
+
+# Alternatives
+
+We could exactly match the Node.js LTS lifecycle. This sets very clear expectations but doesn't give us much freedom to evolve our usage of modern JavaScript features and could possibly force us to maintain our own forks of portions of the JavaScript ecosystem as we would assuredly have one of the longest support policies on record.
+
+# Artifacts
+
+We'll describe this resolution in the README and the documentation for Ember CLI for users.


runspired:
Is there any particular package that keeps us from jumping all the way to the 4.2.x (LTS) release?  I haven't encountered anything not working in the ecosystem thus far and I tend to lock things there.

active/0047-node-support.md
```diff
@@ -0,0 +1,28 @@
...
+
+# Motivation
+
+We've avoided adding libraries like Babel into our ember-cli applications in order to maintain debuggability and reduce magic. This leaves us yearning for the features, flexibility, and syntactic sugar of more-recent V8 releases. In order to not be disruptive we have continued to support the 0.10 branch which has tied us to an ancient V8 engine.
+
+# Proposal
+
+We track the [existing LTS releases for Node.js](https://github.com/nodejs/LTS#lts_schedule), but only provide support for the active LTS releases, or the stable versions which will become LTS releases. Once an LTS is moved into maintenance mode we will drop support for that version. Versions of Ember CLI whose support crosses one of the Node.js boundaries will be supported per the ember-cli release support policy which is currently unspecified, though described by RFC #46.
```
rwjblue: #46 should be #47 I think

stefanpenner:
@runspired consumers of our packages are often tied to specific installations of node. For example ghost (which supports node LTS's), slow moving shops and some mega-corps like Yahoo (although we are nearly off 0.10.x in most cases)

active/0047-node-support.md
rwjblue: Can you spell this sentance out a bit more? Assuming this policy has been in place from the beginning, what would we be supporting today? What will we be supporting after 2017-04-01, etc?

rwjblue:
@erisds / @kevinansfield - I know Ghost has a specific support policy, and I'd love for y'all to review this and see where this lands RE: your own support guarantees...

cibernox:
Maybe I'm biased, but given that ember-cli is tooling we use to build our apps, not something we need at runtime, I feel that the need of maintain support a wide range of versions is low. 
Even if some users needs to use node 0.10 for their backend, they can use a newer version of node with ember-cli using nvm by example.

Therefore, I think that supporting the last 2 mayor versions (now 4.X and 5.X, and in the future 5.X and 6.X) wouldn't be a harsh move.

technomage:
I recommend the active LTS releases

rtablada:
While I love the new and shiny 5.x I think that LTS support may be best.

Knowing what companies like Ghost are supporting will also help with that decision.

Also maybe lining up LTS support for `ember-cli` with the next Node LTS may be an option that makes both sides of things happy?

grapho:
LTS
Less pressure to keep up to date

Gaurav0:
While I do support ensuring 0.12 is supported in the next LTS release, I think it is okay to drop 0.10 support sooner given that it has been deprecated for a long time, so long as we bump the major version number.

Whenever we do drop support for 0.10, we should bump the major version number to indicate a breaking change.

I think it would be premature to drop 0.12 support given that it has not even been deprecated yet. If we want to drop 0.12 support, we should deprecate it very soon, before the next LTS, and drop support in the following LTS.

cibernox:
@grapho Pressure to the developer or pressure to the ember-cli team? Because for the ember-cli team, they are always expected to support the newest and shiniest. For the developer, probably supporting the last 2 versions of Node gives you, given the history of the project, well over 1 year to update. 

@nathanhammond I wonder (I don't use node.js on a daily basis), what kind of features ember-cli wouldn't be able to leverage until April of 2018 because of supporting node 4.2 LTS until end of life (30 months)? As I stated in my previous comment, given that ember-cli falls in the tooling realm and nothing prevents users from having multiple versions of node simultaneously, if narrowing the support matrix pays tangible wins I don't think developers would be pissed by having to upgrade node.js once a year, specially how loudly we complain about non-technical not updating their browsers.

HeroicEric:
Supporting the latest LTS sounds good to me. Another benefit of sticking with the 4.2 LTS for now is that it gives some more time for addons to update to work with npm@3.x, which is bundled with 5.x and still causes some issues.

Gaurav0:
@cibernox For experienced developers, I agree.

But another issue we should consider is that many new developers likely have whatever version of node came with their system. For many systems, including the latest MacOSX, that is still 0.10. Are we sure we want new first time users to have to upgrade node or install nvm just to use ember-cli?

Also, is there any way to transpile on node to get new features?

pixelhandler:
:+1: for current LTS

stefanpenner:
It's also worth noting, the versions of node ember-cli supports, will inherently also enforce that same node version support on add-ons. e.g. If you have an add-on, and want it to be a good member of the ecosystem your add-on should work on all the ember-cli supported versions of node.

Of-course exceptions may exist, but in general ^^ likely stands true.

ErisDS:
Our (Ghost's) version support policy is here: http://support.ghost.org/supported-node-versions/, this includes an explanation of why we support LTS. 

The purpose of Node adding an LTS policy was to create a clear standard for support that the whole community can follow. It states clearly when support for certain version should be dropped and is a clear signal to every part of the community, whether you use Node directly, or Ember with Ember-cli, or you're an add-on or a project using express or Ghost or all of the above. That clarity is what builds reliability, dependability and trust for our community as a whole.

For this reason, I think Ember-cli has a responsibility to support and reinforce the Node.js LTS policy.

stefanpenner:
@ErisDS so you are good with our above proposal ? or..

Gaurav0:
@stefanpenner Based on my reading of the support policy, she is saying we should continue to support 0.10 until October 2016, which is consistent with @nathanhammond 's proposal but not @cibernox 's response.

ErisDS:
> Once an LTS is moved into maintenance mode we will drop support for that version. 

I feel very strongly that this is a misinterpretation of the policy. Support for a node version should end when LTS ends, not when maintenance starts. I believe this needs to be clear and rigid for the benefit of the wider community.

nathanhammond:
@ErisDS @ember-cli/core et al.:

I thought about this a great deal before making my proposal. :smile: From the [Node.js Long-term Support Working Group](https://github.com/nodejs/LTS):

> 1. Once a release enters LTS, no new features may be added to that release. Changes are limited to bug fixes, security updates, possible npm updates, documentation updates, and certain performance improvements that can be demonstrated to not break existing applications, etc. Semver-minor changes are only permitted if required for bug fixes. Semver-major changes are only permitted if required for critical security and bug fixes.
> 2. Once a release moves into Maintenance mode, only critical bugs, critical security fixes, and documentation updates will be permitted.

To clarify: what I'm proposing is our Node.js support for the current `HEAD` of the `master` branch. This is distinct from providing guarantees that some version of Ember CLI will run securely on each Node.js LTS release. We haven't yet proposed our plans for LTS releases of Ember CLI itself (our first LTS is probably six months away). Currently we provide no guarantees of support for _any_ previously released version of Ember CLI–which is a gap we wish to correct and are working toward.

My primary goal with this proposal is to balance the tradeoff of stability vs. stagnation, something the Node.js community has struggled with since its inception. While an existing Node.js LTS release is actively maintained we will be able to easily and gracefully backport fixes to existing Ember CLI releases because `HEAD` will still be compatible with that LTS version. Dropping support when a Node.js release enters maintenance mode for `HEAD` can result in a higher cost for release engineering to backport critical security fixes but doesn't tie the momentum of the project to an almost three year old release of Node.js.

Not having a policy before now we believe that it's important to give people a chance to migrate off of existing versions of Node.js which we currently support (0.10, 0.12) all the way until Node.js completely drops support for them.

Given this strategy I believe that we can make this decision independent of our Ember CLI LTS release policy without worry of future conflict.

nathanhammond:
@runspired @cibernox et al.:

Presently Ember CLI counts as tooling, but we're finding our code being used as the foundation for more and more internal and external to the Ember community. Assuming that we know how everybody wishes to extend Ember CLI would be hubristic of us. With that in mind I believe that it's important to make sure that we err on the side of caution. Yes, this could end up making us wait an additional year for `async`/`await` if it doesn't land in the next Node.js LTS, but being incompatible with the community is a far greater cost than a few language features.

cibernox:
@nathanhammond Just one more question. Being compatible with Node 4.2 means we also need to mantain compatibility with npm 2.X with which it ships by default? Or the supported version of NPM can evolve at a different pace?

I can see the npm 2.X support being more problematic long term than supporting node itself. If being 4.2+ **AND** npm3+ is ok, I see less of problem in supporting the latest LTS until EOL.

nathanhammond:
@cibernox Our `npm` support story is a different question entirely and outside the scope of this RFC. Ember CLI presently bundles its own version of `npm` and we don't see that changing. Suffice to say that we're investigating how to handle package management more gracefully.

cibernox:
> Ember CLI presently bundles its own version of npm

Those are total news to me :exclamation: 
Since when? Running `npm install` from within a ember-cli project is not using the system npm or I'm missing something?

nathanhammond:
@cibernox

> Running `npm install` from within a ember-cli project is not using the system `npm`

`npm install` uses the system version. `ember install` uses our internal version.

stefanpenner:
@cibernox we also are planning to allow `ember install` to work correctly in all cases (even if the local node_modules is partial or non-existent). Before people start worrying, I plan to enumerate the idea in RFC form, so we can gain support/feedback/ideas. 

The issue was, earlier we used the system npm, but navigating N different versions (with there own issues) proved quite troublesome. 

---

This raises a good point, does LTS mean we are stuck with the bundled NPM for the initial ember-cli installation? Or can we assume users can upgrade `npm`, up to this point we have assumed users can upgrade. But maybe this needs fleshing out ..., what are your thoughts on `npm` version @ErisDS ?

nathanhammond:
@stefanpenner

> does LTS mean we are stuck with the bundled NPM for the initial ember-cli installation?

I believe that we need to support whatever the bundled `npm` version is for installation of the global ember-cli (`npm install -g ember-cli`). Beyond that I believe we can take control. We can support only the `HEAD` of each LTS branch, so whatever the most-recent bundled `npm` is.

stefanpenner:
@nathanhammond so as long as ember-cli global is installable by the bundled npm, and we fix `ember install`, we should be in the clear.

cibernox:
@stefanpenner @nathanhammond that sounds right. LTS sounds less of a blocker then.

jaswilli:
I'd love to see ember-cli take a very conservative path when it comes to chasing the latest and greatest, especially if the justification is _wanting_ to use new language features.

Not every developer or project is in complete control of what version of node they can use for all the usual reasons why legacy anything sticks around. If every project was greenfield, or small, or run by a group of likeminded people I'm sure everyone would want to be running the latest stable version of node.

Unfortunately that's not the case, and as soon as the desire to play with new language features wins over the desire to provide a stable platform for the entire community, there are going to be a lot of unhappy ember-cli users out there.

The "Alternative" proposal, in my opinion, is the prudent choice for a tooling/infrastructure project like ember-cli.

stefanpenner:
@jaswilli it isn't just language features, although that is how the conversation started.
- old node, bundle old (questionably working) npm
- old node bundle ancient entirely unsupported versions of libuv and v8, causing real grief.
- some core API and performance are non-existent on some versions (such as execSync or vm on node 10)
- I suspect this list will grow

We will take a conservative stance (as the RFC describes), how conservative is the discussion here. So saying "you should be conservative" isn't really constructive, as that is already a fact we have decided on.

When weighing in the balance, we will aim to be as conservative as practical. I am just glad node 10 will be dropped this year :tada:

jaswilli:
> saying "you should be conservative" isn't really constructive, as that is already a fact we have decided on.

Sorry if I sounded non-constructive, or non-specific. As the last sentence of my post stated, I think doing anything other than supporting the LTS version(s) of node for their entire LTS lifespan is going to be problematic for ember-cli users.

So I'm just weighing in on "how conservative" I think ember-cli should be.

**edit**

> When weighing in the balance, we will aim to be as conservative as practical. I am just glad node 10 will be dropped this year

😄  Yes, there will be much rejoicing.

rtablada:
Not sure that tying to the full LTS length of Node is quite the requirement seeing how their lifecycle is meant for LTS to be built on top of.

So for instance if Node's LTS is 30 months then we shouldn't also be LTS of 30 months since 4.2 would be deprecated by the time ember-cli is EOL.

The first one might be weird since it has to sync up with both Node and Ember Core.

ErisDS:
> So for instance if Node's LTS is 30 months then we shouldn't also be LTS of 30 months since 4.2 would be deprecated by the time ember-cli is EOL.

I'd argue the very opposite. The ceasing of support for a version of Node in dependent projects should cascade FROM the end of LTS not lead up to it. Ember-cli, as a dependent, needs to follow Node's lead.. otherwise this gets super confusing.

>  I am just glad node 10 will be dropped this year :tada:

Me too, for sure!

To explain my perspective in more detail here, what I want is for there to be absolute clarity around what "end of LTS" means, and for the whole community to get behind it. I want to avoid the old versions hanging around indefinitely by setting expectations and creating absolute certainty about when end of life for 0.10 is.

If key dependencies all end 0.10 support at the same time, that sends a very strong and very clear message to everyone from hosts to hobbyists, and from enterprise to entrepreneurs that they must switch away from an end of life LTS by the time the LTS ends.

> To clarify: what I'm proposing is our Node.js support for the current HEAD of the master branch. This is distinct from providing guarantees that some version of Ember CLI will run securely on each Node.js LTS release. We haven't yet proposed our plans for LTS releases of Ember CLI itself (our first LTS is probably six months away). Currently we provide no guarantees of support for any previously released version of Ember CLI–which is a gap we wish to correct and are working toward.

This sounds like something that can be worked with, although I'd echo @jaswilli's point that doing all this just because someone _wants_ to use a latest & greatest version seems like a bad basis for decision making. 

> This raises a good point, does LTS mean we are stuck with the bundled NPM for the initial ember-cli installation? Or can we assume users can upgrade npm, up to this point we have assumed users can upgrade. But maybe this needs fleshing out ..., what are your thoughts on npm version @ErisDS ?

`npm` is something that definitely warrants a different discussion & some fleshing out. The unreliability of older versions is a serious issue - it reflects badly on dependent projects. There doesn't seem to be any leadership on what to do about this at the moment, and I'm keen to find some clarity as we also haven't really covered this in our LTS policy. Not sure if it's possible to find a recommended minimum version based on what's reliable?

nathanhammond:
Regarding @ErisDS's [response](https://github.com/ember-cli/rfcs/pull/47#issuecomment-205974009).

---

> cascade FROM the end of LTS not lead up to it

Agreed, though we may target coincidental deprecation. /cc @rtablada

> This sounds like something that can be worked with...

How should I characterize this? "Tepid assent?"

> ...although I'd echo @jaswilli's point that doing all this just because someone wants to use a latest & greatest version seems like a bad basis for decision making. 

Agreed. @stefanpenner's [response neatly addresses this.](https://github.com/ember-cli/rfcs/pull/47#issuecomment-205952372)

> If key dependencies all end 0.10 support at the same time, that sends a very strong and very clear message to everyone (...) that they must switch away from an (...) LTS by the time the LTS ends.

Agreed, and we can choose to match our release lifecycle and support schedule to Node.js as we become more familiar with our proposed release train model. This is still `undefined` for now.

> `npm` is something that definitely warrants a different discussion & some fleshing out.

We'll post an RFC for discussion on just that topic at some point soon.

---

Another way of thinking about the last 12 months of a Node.js LTS is "life support" mode. We can deal with the cost of writing code to close critical security issues multiple times as that should be presumed infrequent (otherwise we're doing it very wrong). We're effectively saying that you won't get to use the latest and greatest features of Ember CLI on a "maintenance mode" LTS because we won't have an easy way to ship code to you. This doesn't mean that we won't be providing a stable supported already-released version of Ember CLI for that LTS (but details on that will come in a separate RFC), just that no new releases will target that platform.

@ErisDS I'm going to update the RFC to include the clarifications your questioning brought about.

cibernox:
While I see value in supporting all alive LTS versions of node, I'd like to bring another approach that wasn't mentioned in the RFC.

Given that ember-cli gathers analytics about usage, would be reasonable to track the versions of node we use and drop support on a version once its usage falls below some threshold. 

Say, once usage falls below 5% (or LTS support ends, whatever comes first) a deprecation warning is displayed asking people to upgrade, and once it falls below 1% it is removed.

That is how this kind of decisions are usually made when determining the browser matrix in the products we build, and could also be used to determine when the % of people using an ancient version is totally vestigial and it doesn't worth the effort.

nathanhammond:
@cibernox That's a valid approach, but by not attaching a timeline to it makes it difficult for enterprises to plan for migrations. A timeline-based approach allows for planning years in advance. How do we decide when to move if a lot of people are still running an older version? It's entirely possible to get stuck in an IE6 scenario here.

webark:
I think running slightly behind the Node LTS policy is the best bet. Currently, that would mean supporting 0.10 for ember-cli. (this kind of back support shouldn't be as large of a burden moving forward due to the recently released Node LTS policy).

However, if I was to write a new addon, would it be considered "bad form" to write it with features that where not supported prior to the current active LTS (4.x)? And then as I was working on this addon over then next year, cut a branch and a version of my addon that was pegged as a "maintenance" version only  when the node v4 LTS, and subsequently ember cli LTS, gets moved into maintenance mode?

If this was considered "good form", then is there a defined method for alerting users when they do an `ember install` that the version of node they are using is too old, and please downgrade to X version of this addon for continued support. (might be beyond the scope of this RFC)

rwjblue:
`npm` versioning aside, I think that sticking with the Node LTS policy makes this much simpler (and not directly our problem/fault).  This still gets us out of supporting 0.10 in 6 months, and 0.12 in about 12 months.

---

As an aside, I think that what "support" means should likely be called out in this RFC.  Does it mean that the latest stable ember-cli version supports each of the "supported" Node versions? Does it mean that we support the most recent release of each LTS or all releases? Does it mean that we  will provide specific ember-cli versions for each node engine version?  What does it mean for non-LTS versions of Node (i.e. 5.x)?

nathanhammond:
@rwjblue Clarifying everything in an update. :)

cibernox:
@nathanhammond If the TLS ends, the support should end too. That is why I added 

> once usage falls below 5% **(or LTS support ends, whatever comes first)**

I just wanted to highlight a possible scape valve in case real world analytics shows that some LTS version is negligible (0.15%) and yet the commitment initially made will holding back the other 99.85%. Decisions based on analytics seem more "fair".

Raising deprecation warnings with enough time as the usage approaches the EOL threshold point might give enough visibility for the enterprise world.

Just wanted to propose an option C

webark:
@cibernox one issue with basing things off of usage that I've noticed in the browser world, is that you can start dividing people up into the "big dogs" and the "small fry". Meaning.. if 0.15% of your users are just random people, it's fine cutting them out, but if 0.15% is a larger corporation, then that puts them in a more "privileged" status. If you just adhere to the formal support policies, it creates more of an even playing field.

_(was wanting to add a Houyhnhnm reference in there, but couldn't find a good fit ;)_

nathanhammond:
[I've updated the RFC per all of y'all's feedback.](https://github.com/ember-cli/rfcs/pull/47/commits/03523311c5b59098b0554c07b44e035a0e6a1388) [Here it is rendered.](https://github.com/nathanhammond/ember-cli-rfcs/blob/node-support/active/0047-node-support.md)

No real changes, but examples have been added, clarifications included.

ErisDS:
> > This sounds like something that can be worked with...
> 
> How should I characterize this? "Tepid assent?"

My concern here is brought to light by one of @rwjblue's questions:

> Does it mean that we will provide specific ember-cli versions for each node engine version? 

As a build tool, ember-cli exists to be a dependency of other projects. However, if you end up not supporting ALL the LTS versions in one single version, then none of your dependents can realistically do this either. The worry is that at some point ember-cli will grandfather 0.10 support in such a way that would force Ghost to choose between 0.10 support or later LTS support meaning there'd be no technical way for us to stick to the policy that we set out ~6months ago.

If we have to hold back and stay on an older version of ember-cli that at least works for all the LTS versions we promise to support then that's at least a workable solution, but not a preferable one. We do our best to keep up with you guys and would rather not be excluded :wink:

nathanhammond:
@ErisDS This RFC does indeed propose an approach that drops the ability to support all of the Node.js LTS versions using the `master` version of Ember CLI. It does not drop the ability to support all of the Node.js LTS versions using some combination of guaranteed supported (LTS) releases and our current `master`. Presumably as well the older releases may "just work" in more-recent Node.js releases. Any support at all for older Node.js LTS releases is already remarkably one-sided [as evidenced by the reactions here](https://github.com/ember-cli/rfcs/pull/47#issuecomment-205589412).

Landing of this RFC would mean that Ghost would adopt Ember CLI's LTS releases, same as you've adopted Node.js's LTS releases. This is a tradeoff you've already elected to make and the reasons you describe apply the same here.

I'm of the opinion that maintaining full feature support for a maintenance mode Node.js LTS release is beyond even the guarantees that Node.js provides. This RFC's proposal feels in line with the spirit of the Node.js release cycle in which the oldest LTS is in maintenance mode. It's supported, but you get no new features. I also have significant concerns about the longevity of our dependencies supporting the oldest of Node.js versions and we can't afford to fork and maintain an entire ecosystem (we already own too much).

ErisDS:
Note: I'm about to get on a plane and so this is the last reply I can write for ~12 hours.

> Landing of this RFC would mean that Ghost would adopt Ember CLI's LTS releases, same as you've adopted Node.js's LTS releases. This is a tradeoff you've already elected to make and the reasons you describe apply the same here.

This might be true in theory, but in practice they are quite different.

One version of Ghost is able to support any version of Node.js. There is no configuration necessary in Ghost itself to make this possible - a user installs Node.js on their environment and Ghost master will work. 

One version of Ghost is not able to support multiple versions of ember-cli. The fact that we depend on you through other mechanisms means we have to pin our support to a version, or range of versions via configuration. There is no way to switch ember-cli version based on a user having a different version of node. 

What you're proposing is rendering any dependent project of ember-cli **incapable** of supporting Node.js's current LTS plan. 

Obviously, that's not OK?

nathanhammond:
@ErisDS Have a great flight, I'm really looking forward to continuing this conversation when you return. :smile:

rwjblue:
@nathanhammond - The most recent round of updates definitely adds some much needed flesh to the proposal. 

> ... will provide support for any Active (not Maintenance Mode) Node.js LTS and the current stable Node.js version(s)

I still believe (as mentioned earlier in this thread) that we should drop this caveat, and support all supported Node LTS versions.

> @ErisDS - My concern here is brought to light by one of @rwjblue's questions:
> 
> > Does it mean that we will provide specific ember-cli versions for each node engine version?

This is also my concern. I see very little value to drop an LTS during its maintenance mode.  The point of maintenance mode is to provide overlap with new LTS as a mechanism to migrate.

jaswilli:
> This RFC's proposal feels in line with the spirit of the Node.js release cycle in which the oldest LTS is in maintenance mode. It's supported, but you get no new features.

Is it? Here's why I think not supporting the entire LTS lifetime is problematic (and please tell me if I'm misunderstanding the situation):

LTS `v.something` is released and people adopt it and plan on having 30 months (or whatever) with this particular LTS train. No new features is fine but bug and security fixes are still coming and (to them) that's all that matters. Now let's say that 18 months in, LTS `v.somethingNew` is released with a version of V8 that supports the new javascript syntax du jour. Support for `v.something` is dropped and [insert new language feature] makes its way into ember-cli.

Now ember-cli  is _broken_  for LTS `v.something` and there's no "best effort" or "should probably still work" about it. All of a sudden that "we've got LTS stability for 30 months" bill of goods I sold my team/users/whomever is no good.

nathanhammond:
> What you're proposing is rendering any dependent project of ember-cli incapable of supporting Node.js's current LTS plan. - @ErisDS

It depends on how you interpret Node.js's LTS plan (more below). I'm proposing that a user who is brand new installing Ghost would need to be aware of their Node.js version and install a packaged release of Ghost that supports that particular version as opposed to the `latest` tag on `npm`. Assuming that a user does that at the beginning of an LTS cycle their experience is unchanged and their application will run, securely, for the entire lifecycle of the LTS (36 months, including maintenance). After 24 months they'll stop receiving new features but the application will run securely and perfectly fine. At the tail end of the Maintenance period they can update their applications, achieving the no-churn thirty month promise.

I'm in effect proposing that ember-cli and Ghost adopt a support policy matching Node.js's including the lifecycle phases.

**Note that Ember itself doesn't provide a guarantee of support for this long (caps at ~15 months for security releases).** This is actually the primary limiting factor for Ghost upgrades right now, not Ember CLI.

---

Regarding whether or not it is in spirit with Node.js's release process, here is Node.js's description.

During Active Support:

> Changes are limited to bug fixes, security updates, possible npm updates, documentation updates, and certain performance improvements that can be demonstrated to not break existing applications, etc. Semver-minor changes are only permitted if required for bug fixes. Semver-major changes are only permitted if required for critical security and bug fixes.

During Maintenance:

> only **_critical**_ bugs, **_critical**_ security fixes, and documentation updates will be permitted.

In short, the Node.js team gets to effectively stop development on their LTS branch after 24 months and only backport **_critical**_ security fixes. During the last 18 months of that period they land only bugfixes, with very limited actual development on that branch.

This proposal provides broader support than Node.js's own proposal as it guarantees:
- Full feature support for the entire Stable => LTS lifecycle (24 months).
- (Presumably) a parallel supported LTS version supporting the oldest maintenance mode release. (The trailing 12 months.)

---

> I see very little value to drop an LTS during its maintenance mode. - @rwjblue

Maintenance mode is an additional twelve months. We'll be supporting a 36 month (or older!) version of V8 at that point, or described differently, a V8 engine from 6 JavaScript library half-lives ago. This extends our support lifetime to 4x beyond what even Ember guarantees.

> The point of maintenance mode is to provide overlap with new LTS as a mechanism to migrate. - @rwjblue 

There is already a six month overlap period every 18 months, even excluding Maintenance mode timeframes.

I only propose dropping support for running `master`. I propose that it use one of our existing still-supported releases, likely an LTS release, likely one which was already intentionally adopted by our consumers. There is no reasonable use case in which a person starts a new project on an old Node.js version so the only people who find themselves in this scenario will simply have to upgrade their Ember CLI version at the same time they're updating their Node version.

I also _really_ don't want to get into the business of dealing with the eventual ecosystem drift wherein we slowly own more and more because of people using new features in our dependencies.

mfeckie:
I think it's pretty reasonable to only 'officially' support the LTS versions node.  I've been using Ember-CLI since it started and have used it with node and iojs, old versions and the bleeding edge.  Have had almost zero issues due to node/iojs version.  Compared to the hell that was bower versioning this seems like less of a problem.

In an exploratory sense, are there things in Ember-CLI / on the horizon that explicitly require a specific node version??

Gaurav0:
@nathanhammond Thanks for the update and for bringing this RFC forward! However, I looked at the recent commit and it does not address the concerns I still have regarding:
1. How soon shall we deprecate (throw a warning for users running) node 0.12?
2. Semver requires that we bump a major version number when dropping support for a node version.
3. What about node / io.js versions 1, 2, and 3? These should be explicitly mentioned.

To be clear, I am in general agreement with the timeframes and reasoning presented in the RFC.

nathanhammond:
@Gaurav0 I address each existing supported version of Node.js here:
- [Legacy](https://github.com/ember-cli/rfcs/pull/47/files#diff-cb989872d0129a613d80eb8ba911d8c3R23)
- [Current](https://github.com/ember-cli/rfcs/pull/47/files#diff-cb989872d0129a613d80eb8ba911d8c3R28)

No other releases are presently supported, or tested against Ember CLI.

Deprecation notices should go in now as 0.10 (already there) and 0.12 will definitely have their support dropped on 2016-10-01 and 2017-04-01 regardless of this RFC as that is when they are retired from support by Node.js itself. Would you be willing to submit a PR to update our current `node` version check?

SemVer is for some combination of humans and machines and doesn't by my estimation _require_ anything with regards to major version bumps. [This section of the spec provides guidance](http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api) and is focused on the human and interface aspect. The decision to increment a major version number is one we'll approach with an analysis of how it impacts the community.

Gaurav0:
Thanks @nathanhammond for clearing things up.

I would simply suggest that a simple statement saying node.js / io.js versions 1, 2, and 3 are unsupported be added to the RFC.

I disagree regarding SemVer. I point to Section 8: http://semver.org/#spec-item-8 The way I read it, a SemVer compliant package including a backward incompatible change to public api MUST increment the major version number

The question becomes whether the support for node v0.10 and v0.12 are part of our public api. Because the documentation no longer specifies this (and has not since ember-cli 2.0) I think it is currently not a part of our public api. I think it should be, as this is critical information for many users.

Since we have agreed lockstep versioning is not strictly required, I think it is important we consider following SemVer strictly. I think we could argue not to have a major version bump for 0.10 because it is already deprecated (although our warning doesn't use that word). However, a think a major version bump for removing support for 0.12, and all future versions of node that we drop support for, should be seriously considered as it removes confusion.

I think we are planning to bump major version around or before 2017-04-01 anyway?

Anyway, thanks for listening to my concerns and being willing to address them, however you decide.

rwjblue:
> I disagree regarding SemVer. I point to Section 8: http://semver.org/#spec-item-8 The way I read it, a SemVer compliant package including a backward incompatible change to public api MUST increment the major version number.

FWIW, our public API will not be changing at all (no API's will be added or removed to drop support for a given node engine) so I'm not certain that this section actually applies here.

> I think we are planning to bump major version around or before 2017-04-01 anyway?

I don't recall having made this decision.  It may absolutely come to pass, but it certainly isn't a thing we have already decided.

---

That being said, I think dropping engine version support via a major version definitely makes sense.  But I'm not sure we are _required_ to do it, and as such I'm not sure that we should call it out in this document.

musaffa:
Regarding the language features

In the battle between `Node` vs `Babel`, I'm supportive of Node for its speed, debuggability and less magic.

Node's [ES6 features](https://nodejs.org/en/docs/es6/) come in three groups: `shipping`, `staged`, `in progress`. Ember CLI can detect the Node version and cherry pick the `Babel` features which are not yet supported by that Node version. Ember CLI can maintain compatibility configs for LTS and stable versions. This way anyone who wants to use the latest Node features will depend less on Babel in their development environment. On the other hand, those who wants to stay with older Node versions can happily use them.

I propose a separate add-on for `node-babel-compatibility` whose sole purpose will be to track ES6 features of node versions and cherry pick the absent features from Babel. The add-on can also track Node `head` and support `staged` and `in progress` ES6 features of released Node versions for the adventurous ember developers who are eager to test them.

Things to consider: How to resolve V8 version incompatibility between Node and PhantomJs during testing?

stefanpenner:
downside of transpiling:
1. transpiling on the fly will hurt boot times
2. debugging is harder, without better baked sourcemap support in node.

rwjblue:
We could easily make it so that you can use the git repo directly for @nathanhammond's proposed "non-maintenance mode LTS" versions, but allow usage within all of them via a prepublish step.  This solves both of the issues @stefanpenner mentioned:

1) Transpiling on the fly is slow -- We wouldn't be doing this
2) debugging is harder because sourcemaps suck (paraphrasing) -- The files on disk in node_modules **exactly** match what is being ran...

musaffa:
Testing in ember isn't quite pleasant enough - may be because of transpiling workflow.

Edit: A new testing workflow proposal may go into another `rfc`.

stefanpenner:
@musaffa that comment is entirely unrelated to this thread, let me suggested you remove it and open it in a fresh thread.

musaffa:
@stefanpenner done. :)

stefanpenner:
> We could easily make it so that you can use the git repo directly for @nathanhammond's proposed "non-maintenance mode LTS" versions, but allow usage within all of them via a prepublish step. This solves both of the issues @stefanpenner mentioned:
> 
> 1) Transpiling on the fly is slow -- We wouldn't be doing this
> 2) debugging is harder because sourcemaps suck (paraphrasing) -- The files on disk in node_modules exactly match what is being ran...

This all seems complicated and doesn't take into account add-ons nor does it take into account npm dependencies that may or may not be transpilable.

kellyselden:
I would like to bring up an idea we briefly touched on at our last face-to-face. Bundling node with ember-cli. Very similar to how we currently bundle npm. We become in control of whatever node version we want, which would be a balance between stability and most recent.

Downsides:
- install size increases (may be a problem for Nitrous users and other container based solutions)
- don't get to use the the system installed if that was your desire

Upsides:
- codebase only has to consider one version of node
- newer v8 (or chakra) features are usable way sooner

I can't offer much in technical challenge of doing this, maybe someone who has made an electron app or otherwise bundled node can comment. If someone smarter than me wants to tell me this idea is nonsense, that is ok too :smile:

musaffa:
> We could easily make it so that you can use the git repo directly for @nathanhammond's proposed "non-maintenance mode LTS" versions

schedule

Here we can see there are at least two overlapping "non-maintenance model LTS" versions. If we support both of them as well as the stable versions according to @nathanhammond proposal, then we cannot use new features introduced in the latter LTS and stable versions unless the former LTS goes into the maintenance stage. A stable version with new features is released just after the LTS version moves into the active LTS stage. This active LTS stage is 18 months long. This means we will be 18 months behind before supporting the new features. If the proposal' aim is to get the new features as early as possible in the master branch, 18 months doesn't sound much early.

nathanhammond:
@musaffa This proposal results in adopting versions of Node.js one year earlier than we would otherwise. This cuts our lifetime of support for particular versions of Node.js to 24 months from 36 months, reducing it to only 4 JavaScript library half-lives.

rwjblue:
> reducing it to only 4 JavaScript library half-lives

Obviously, this is not a thing (and doesn't really contribute to the discussion) :smiling_imp:

musaffa:
@nathanhammond Thinking the other way, supporting all the LTS will require 30 months before adopting the new features. The current proposal will reduce it to 18 months. This is definitely better still it isn't early adoption (1.5 years!). As we are going to the special route of `master` branch, can we make the early adoption time gap even smaller?

rwjblue:
> As we are going to the special route of master branch, can we make the early adoption time gap even smaller?

As I stated earlier, I see little value in this.

musaffa:
@rwjblue I may be wrong in thinking that the early adoption of node's ES6 features means less dependency on transpilation.

rwjblue:
@musaffa - We do not currently transpile anything that runs in Node (only things for the browser), and this proposal (as written) does not suggest adding Node-land transpilation.  I suggested it in https://github.com/ember-cli/rfcs/pull/47#issuecomment-207053537 as a possiblity, but @stefanpenner seems fairly opposed (see https://github.com/ember-cli/rfcs/pull/47#issuecomment-207138117).

rwjblue:
@stefanpenner:

> doesn't take into account add-ons

Ember CLI can definitely **not** force the entire addon ecosystem to follow the same policy (no matter what it is that we decide here).

>  nor does it take into account npm dependencies that may or may not be transpilable.

Ensuring that our direct dependencies (and those of a default application) meet our requirements is an issue with or without transpilation.

kategengler:
Having read through this thread, I really like the idea of bundling node as @kellyselden mentioned. 
- Doing so should mean that new users would not be tripped up by whatever version of node is on their system. 
- We would always know the node version used for any version of ember-cli. (Potentially easier issue triage & debugging). 
- It would also mean that addon authors wouldn't need to test against a matrix of (node versions) x (ember-cli versions) x (ember versions), just (ember-cli versions) x (ember versions), making it easier to create and maintain an addon that supports multiple ember-cli versions.
- No users would ever have to upgrade Node on their system due to a change in ember-cli.

stefanpenner:
Although I also like the idea of bundling our own node, I do believe that it may not be aligned with our ability to support projects who depend on ember-cli as a lib. That is, unless they also want to adopt bundling a node. Which I suspect they do not.

I suspect the bundling of our own node, will most likely be limited to GUI installer based workflows, hearth etc. but for libraries like ghost or other consumers I suspect that may be DOA. Although would welcome others input.

@rwjblue although we can't enforce add-on node version support, by virtue of ember-cli having a conservative support strategy, we will implicit imply that to the larger ecosystem.

Also, if someone can demonstrate that the transpiled versions for legacy support won't be at high maintenance/debugging cost to us. And that it will somehow work with our dependencies (eg. I don't believe we want to support multiple working trees). I can be convinced. My current stance is purely pragmatic. Aiming to minimize cost and confusion.

nathanhammond:
Today one of our dependencies dropped Node.js 0.10 support. The [actual commit that broke it is this one](https://github.com/jprichardson/node-fs-extra/commit/cfd0fa25f05a185a0758a3cb12e52854cc3170e4), just two commits after the [note of dropping formal support here](https://github.com/jprichardson/node-fs-extra/commit/2d2758e2fd6bdbff280cd8e443e020485d8b82c5).

This is six months prior to the official end of support for 0.10. node-fs-extra is a pinned dependency for us so it doesn't have an immediate impact, but we are now assured to be running an out-of-date version of this until we drop Node.js 0.10 support.

At this point I'm unsure how common this will be, but we should take particular care to note what is happening in the ecosystem.

jprichardson:
> just two commits after the note of dropping formal support here.

The note was posted in Jan. Stating "just two commits" feels a bit disingenuous as if I had just dropped support after I had announced dropping support.

Regardless, I'm happy to bring it back official support if it's critical / very important. See rest of comments here: https://github.com/ember-cli/ember-cli/pull/5853#issuecomment-215445275 Let me know your thoughts.

jprichardson:
Brought official support back for Node v0.10 in fs-extra. https://github.com/ember-cli/ember-cli/pull/5853#issuecomment-215452718 Like you, I'll end support for Node v0.10 on 2016-10-01.

nathanhammond:
@jprichardson I didn't mean to trivialize the amount of time between your decisions, I'm sorry that the words I chose implied any recklessness on your part. That wasn't my intent. :smile:

We too are eagerly counting down the days until we can drop v0.10 and v0.12 support. The part that is instructive for us (@ember-cli/core) to learn from is that for a project as large as ours there is no such thing as unofficial support. The moment we stop testing something we will find ourselves introducing a change that breaks support for that version.

We're grateful that you're willing to support v0.10 with us until 2016-10-01. :smile:

stefanpenner:
> We're grateful that you're willing to support v0.10 with us until 2016-10-01.

:+1:

@jprichardson i would be curious in hearing your thoughts on the rest of our proposal. As it will most likely affect other maintainers such as your self.

nathanhammond:
Closing as we are deferring to the default of the Node.js ecosystem. Our Node support policy at present is simple:

> Ember and Ember CLI have committed to supporting the [Node.js LTS schedule](https://github.com/nodejs/LTS#lts-schedule) for the `HEAD` of our `master` branch(es). This means that we will will drop support for 0.10 on 2016-10-01 and 0.12 on 2016-12-31 per the Node.js Long-term Support Working Group's schedule.

Thank you all for your thoughts. Blog post forthcoming.

PR #48, opened 2016-04-05
Rename master to canary.

nathanhammond:
[Rendered](https://github.com/nathanhammond/ember-cli-rfcs/blob/master-canary/active/0048-master-canary.md)

We recognize that renaming `master` to `canary` is contentious. We're posting this for short-form feedback. **Please respond using reactions on this post!** We'll use that to gauge community interest, and engage in conversation on this if it seems like there is significant momentum in one direction or another.

### DIFF
--- a/active/0048-master-canary.md
+++ b/active/0048-master-canary.md
@@ -1,0 +1,18 @@
+- Start Date: 2016-04-04
+- RFC PR: #48
+
+# Summary
+
+Rename the `master` branch `canary`. We recognize that this is contentious and are seeking community feedback.
+
+# Motivation
+
+The goal is to make it a bit more clear which channel you're on without having to mirror commits between `master` and a `canary` branch.
+
+# Drawbacks
+
+This makes it a bit more confusing for new contributors to the project, but is neatly handled by GitHub's interface.
+
+# Alternatives
+
+Do nothing.


rwjblue:
I am in favor of this. I feel that when we adopt the channel based release process (specified in #46 I believe) this will make it much clearer to users what channel they are using and will remove a fairly large number of WAT's that we see in Ember itself when describing which branches track which channels.

Turbo87:
FWIW I prefer master/beta/stable or master/2.6.x/2.5.x

kellyselden:
After thinking about this, I think I prefer to delay this discussion/implementation until after the release channels go out. Then we will have a way better idea if master vs. canary is an issue or non-issue.

rwjblue:
Hmm, I'm not sure I agree.  I believe delaying puts us in a place where we have to decide to stop supporting people that are using `#master` explicitly. It seems better to me to decide this, and do a single "documentation push" to explain the concepts....

nathanhammond:
We're at 50/50 support, serving as a valuable referendum on the community. We'll discuss this in person next week.

Turbo87:
in last weeks team meeting it was decided to stay at `master` for now and create a `canary` branch or tag that automatically tracks the `master` branch.

we don't know how exactly that tracking thing should work yet though and are looking for implementations that can create such a mirror-branch.

rwjblue:
Right, so closing this RFC...

PR #49, opened 2016-04-05
Add method for addons to know their parent type

zackthehuman:
[Rendered](https://github.com/zackthehuman/ember-cli-rfcs/blob/belongs-to-addon/active/0049-belongs-to-addon.md)

Sometimes addons need to know if they're included within an app or another addon. This proposes adding a method that allows an addon to easily know this fact.

### DIFF
--- a/active/0049-belongs-to-addon.md
+++ b/active/0049-belongs-to-addon.md
@@ -1,0 +1,40 @@
+- Start Date: 2016-04-05
+- RFC PR: #49
+
+# Summary
+
+Add a hook that allows Ember Addons to know whether they're being included in an app or inside of another addon.
+
+# Motivation
+
+Some addons may need to do extra or less tasks when being the child of an addon versus an app. Currently there is not a documented or prescribed way to determine if an addon's parent object is an app or another addon. Since there are varying ways to determine this, it seems like a good idea to provide a standard method to call that resolves this question.
+
+# Detailed design
+
+We could expose a method on `EmberAddon` that returns a boolean that indicates whether its parent is an app or an addon.
+
+```javascript
+belongsToAddon: function() {
+  return flag; //... true if the parent is an addon, false otherwise ...
+}
+```
+
+Alternatively it could return a string, more or less a kind of enumeration, that describes the parent's type.
+
+```javascript
+parentType: function() {
+  if (parent type is an app) {
+    return 'app';
+  } else {
+    return 'addon';
+  }
+}
+```
+
+# Drawbacks
+
+This may be a slight edge case for Ember addons out in the wild. However, internally at my company there are addons that include other addons frequently, which makes this use case very common for me.
+
+# Alternatives
+
+Do nothing. Let everyone determine their parent type for themselves.


runspired:
In theory, I love this. In practice, I suspect this just puts a new coat of paint on a very bad scope leak problem for the addons which hit it.

**Scenario 1:**
Addon `A` consumes addon `B`.
`App` consumes `A`.

Currently, code from `[B]/app/` will be included in `App`.  Smart addons could use this proposed flag to maybe filter out some things from being automatically exposed.

**Scenario 2:**
Addon `A` consumes addon `Ba`.
`App` consumes `A` and `Bb`.

Currently, if `Ba` is a dependency of `A`, then (usually based on alphabetical order) `App` may end up with `Ba` as well instead of it's declared dependency of `Bb`.  This particular scenario is (I think) why addons within addons (vs as peerDependencies) isn't something that's been explicitly supported to this point, until we can package them correctly, this introduces difficult to debug problems.

If we add this API, we may find that our newly explicit support of addons having other addons as dependencies amplifies this problem in userland.

zackthehuman:
@runspired I completely agree.

However, the addons within addons issue exists regardless. Currently, since there is no API to use in this case, different developers do different things. For example, some may check if `this.app.app` is a thing, some may inspect `this.parent.parent`, there are untold other ways to divine this information.

From my perspective there is still value in providing a "standard" way to do it even if it's not something we want to promote as a best practice. At least an API gives us a hook that we can deprecate in the future -- when we've vetted out how addon-addons are _supposed_ to work.

A concrete example of when this would be useful is if your addon needs to `app.import` something. It's not possible to do this if the parent object is another addon, because it doesn't have an `import` method. If this scenario was easy to detect, you could handle it nicely by emitting a warning message or by "finding" the actual `app` object and `import`ing, or doing nothing at all because it may only make sense to import within an `app` and not within an `addon` context, so you instruct the developer to include the addon in the consuming application.

runspired:
Most the current methods (and even to a degree this) still require you to "bust through" to find the app when needed.  Maybe A is an addon for B which is an addon for C which is an addon for you App.  I'm sometimes surprised I haven't seen this happen yet:

```
while(app && app.app && !app.import) {
  app = app.app;
}

if (!app.import) {
  throw new Error("This addon can't find the parent app!");
}
```

trabus:
While I've tried to avoid such situations, I can understand the need to know where an addon is in the project stack. I'd prefer to provide an official API rather than require addon developers to come up with various solutions that may or may not break in the future.

So in general I am :+1: on this.

active/0049-belongs-to-addon.md
```diff
@@ -0,0 +1,40 @@
+- Start Date: 2016-04-05
+- RFC PR: #49
+
+# Summary
+
+Add a hook that allows Ember Addons to know whether they're being included in an app or inside of another addon.
+
+# Motivation
+
+Some addons may need to do extra or less tasks when being the child of an addon versus an app. Currently there is not a documented or prescribed way to determine if an addon's parent object is an app or another addon. Since there are varying ways to determine this, it seems like a good idea to provide a standard method to call that resolves this question.
```
stefanpenner, line 10: can examples be provided?

active/0049-belongs-to-addon.md
```diff
@@ -0,0 +1,40 @@
...
+}
+```
+
+# Drawbacks
+
+This may be a slight edge case for Ember addons out in the wild. However, internally at my company there are addons that include other addons frequently, which makes this use case very common for me.
+
+# Alternatives
+
+Do nothing. Let everyone determine their parent type for themselves.
```
stefanpenner, line 40: Add-on hierarchies are complicated, especially since `npm` dupes, but our UI may not always want dupes. An alternative is to explore how this problem can be further specified, and then addressed.

stefanpenner:
The current  proposal seems to explore solving a symptom, the symptom is that of the tension between non-duped client-side deps, vs duped NPM deps.

I believe several very short case-studies of addons that fall into this should be added to this RFC so we can speak concretely about the problem(s). That will most likely enable proposal the best possible solution or series of solutions to tackle the root cause(s).

notmessenger:
https://github.com/softlayer/ember-cli-jsdoc/blob/master/blueprints/ember-cli-jsdoc/index.js#L144-L149 and https://github.com/softlayer/sl-ember-components/blob/master/index.js#L59-L68 are concrete examples of such usage and need, as requested by @stefanpenner

rwjblue:

Thank you!

PR #52, opened 2016-04-23
Adds support to outputFile for addons JS

MiguelMadero:
[rendered](https://github.com/MiguelMadero/rfcs/blob/addon-output-file/active/0000-app-import-output-file-for-addons.md)

cc: @stefanpenner @nathanhammond

### DIFF
--- a/active/0000-app-import-output-file-for-addons.md
+++ b/active/0000-app-import-output-file-for-addons.md
@@ -66,6 +66,15 @@ app.import('addons/ember-addon-2.js', { outputFile: 'assets/alternate-vendor.js'
 app.import('addons/ember-addon-n.js', { outputFile: 'assets/alternate-vendor.js'});
 ```
 
+## Trees
+
+A typical addon have different trees with JS that are relevant for this discussion. For a `app.import('addons/ember-lgtm', { outputFile: 'assets/vendor-deferred.js'});` we will do the following:
+
+* `app` will still be added to app.js
+* `test-support` will still be part of the app's test tree.
+* `addon` and `vendor` after this is implemented, it will be concatenated with other content in `vendor-deferred.js`. Note that an addon's `vendor` tree refers to `app.import`s done from the addon's `index.js`, not the `vendor` folder.
+
+
 # Drawbacks
 
 * Potential confusion about static and "addon" files. It isn't clear which ones are procesed and which ones are not.
--- a/active/0000-app-import-output-file-for-addons.md
+++ b/active/0000-app-import-output-file-for-addons.md
@@ -1,0 +1,116 @@
+- Start Date: 2016-04-03
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+Allows `app.import` to specify outputFile for addons.
+Not calling `app.import` for an addon would be considered to have
+an outputFile of `assets/vendor.js` to maintain the existing behavior.
+
+# Motivation
+
+[RFC#28](https://github.com/ember-cli/rfcs/pull/28) added support to specifiy an `outputFile` this works great for static `vendor` files, however, the addon's processed JS always end up in `vendor.js`. We have the same motiviation to do this for addons as we do for other vendor assets. This is specially important for large applications where only parts of the app depend on certain addons. Splitting the output vendor files allows us to cache assets more efficiently or even lazy-load them. Note: lazy-load is out of the scope of this RFC
+
+
+# Detailed design
+
+Today the addonTree processes JS, which get contenated together to `addons.js` and later to `vendor.css` and `vendor.js` respectively. With this change, we will allow the addon consumers further control about where the processed JS output for each addon ends up. Today we don't process addon's CSS directly this way, but we already have a solution via `outputPaths`, for this proposal CSS is ignored. See alternatives for a consistent option with CSS and JS.
+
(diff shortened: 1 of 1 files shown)


active/0000-app-import-output-file-for-addons.md
```diff
@@ -0,0 +1,116 @@
...
+
+# Summary
+
+Allows `app.import` to specify outputFile for addons.
+Not calling `app.import` for an addon would be considered to have
+an outputFile of `assets/vendor.js` to maintain the existing behavior.
+
+# Motivation
+
+[RFC#28](https://github.com/ember-cli/rfcs/pull/28) added support to specifiy an `outputFile` this works great for static `vendor` files, however, the addon's processed JS always end up in `vendor.js`. We have the same motiviation to do this for addons as we do for other vendor assets. This is specially important for large applications where only parts of the app depend on certain addons. Splitting the output vendor files allows us to cache assets more efficiently or even lazy-load them. Note: lazy-load is out of the scope of this RFC
```
stefanpenner, line 13: So the idea is as an escape valve for manually splitting vendor, rather then working towards doing it more automatically? Seems reasonable, as long as it doesn't cause the more automatic approach grief.

cc @chadhietala would love your eyes on this.

active/0000-app-import-output-file-for-addons.md
```diff
@@ -0,0 +1,116 @@
...
+app.import('addons/ember-addon-1.js', { outputFile: 'assets/alternate-vendor.js'});
+app.import('addons/ember-addon-2.js', { outputFile: 'assets/alternate-vendor.js'});
+app.import('addons/ember-addon-n.js', { outputFile: 'assets/alternate-vendor.js'});
+```
+
+# Drawbacks
+
+* Potential confusion about static and "addon" files. It isn't clear which ones are procesed and which ones are not.
+* Potential confusion about using `addons/` as the path.
+* Potential confusion about the order since addons not specified are still "imported" into vendor.js and vendor.css
```
stefanpenner, line 82: - which parts of the app depend on which add-ons will prove to be confusing, how will this be addressed?
  - without a proper dep-graph this seems rather haphazard. Which makes it somewhat risky to provide a high level api for.
- what happens if an add-on has a `app` directory, do we detect that and error if they try to `app.import` ?

active/0000-app-import-output-file-for-addons.md
```diff
@@ -0,0 +1,116 @@
...
+an outputFile of `assets/vendor.js` to maintain the existing behavior.
+
+# Motivation
+
+[RFC#28](https://github.com/ember-cli/rfcs/pull/28) added support to specifiy an `outputFile` this works great for static `vendor` files, however, the addon's processed JS always end up in `vendor.js`. We have the same motiviation to do this for addons as we do for other vendor assets. This is specially important for large applications where only parts of the app depend on certain addons. Splitting the output vendor files allows us to cache assets more efficiently or even lazy-load them. Note: lazy-load is out of the scope of this RFC
+
+
+# Detailed design
+
+Today the addonTree processes JS, which get contenated together to `addons.js` and later to `vendor.css` and `vendor.js` respectively. With this change, we will allow the addon consumers further control about where the processed JS output for each addon ends up. Today we don't process addon's CSS directly this way, but we already have a solution via `outputPaths`, for this proposal CSS is ignored. See alternatives for a consistent option with CSS and JS.
```
stefanpenner, line 18: What happens to modules in `addon-path/app`? Are they left in app.js, and error if loaded before the addon is loaded or... ?

Are all trees of an add-on moved into these new files, or are some not? For example, lets look at https://github.com/mike-north/ember-cli-materialize/tree/master/app  vs https://github.com/mike-north/ember-cli-materialize/tree/master/addon vs https://github.com/mike-north/ember-cli-materialize/tree/master/vendor

active/0000-app-import-output-file-for-addons.md
MiguelMadero, line 13: > So the idea is as an escape valve for manually splitting vendor

Yes. 

I would like to learn more about the plans for the automatic approach.

active/0000-app-import-output-file-for-addons.md
MiguelMadero, line 18: I was thinking that the `app` tree still goes to `app.js` and the addon and vendor tree will go to the specified outputFile. The error would be no different from what we have for static vendor assets if they are not loaded.

stefanpenner, line 18: ya, it seems like @chadhietala work merely add extra sanity to this. So it could be layered on after (at-least it seems like it).

active/0000-app-import-output-file-for-addons.md
MiguelMadero, line 82: > which parts of the app depend on which add-ons will prove to be confusing, how will this be addressed?

I don't think this is any different from what we have for static vendor assets. I think a proper dep-graph could certainly help, but I'm not sure it is riskier than what we can already do today. 

> what happens if an add-on has a app directory, do we detect that and error if they try to app.import ?

I think that throwing would be simpler. However, if the addon's app.import type is for `vendor` we can add it to the same `outputFile`, and maintain the current behavior for type `test`. The problem with adding it to outputFile is when we have more than one addon that imports the same file. Today we'll add the content only once, but if the outputFile is different I think we should at least warn that both `outputFile`s will contain the same content.

active/0000-app-import-output-file-for-addons.md
MiguelMadero, line 18: I added more info about this on a "Trees" section under Detailed Designed.

rwjblue:

Thank you!

PR #53, opened 2016-05-04
Autopick port & keep track of an app being served

sivakumar-kailasam:
[Rendered](https://github.com/sivakumar-kailasam/rfcs/blob/autopick-available-port/active/0000-autopick-available-port.md)

### DIFF
--- a/active/0000-autopick-available-port.md
+++ b/active/0000-autopick-available-port.md
@@ -18,7 +18,7 @@ This RFC is a followup to [ember-cli/ember-cli#5703](https://github.com/ember-cl
 - In case the user tries to run `ember serve` on an app which is already running, we will use the port number saved in the file to let the user know the entire url where the app is already being served
 
 #### Picking a random port when the default is unavailable
-- If port `4200` is already being used, we can find the next available port. This will be similar to the live-reload-port's existing implementation.
+- If port `4200` is already being used, we can find an available port in the unassigned range of 49152–65535. This will be similar to the live-reload-port's existing implementation.
 - There will be an extra message printed before we print the url during initial build informing the user that the app was launched in a different port. This can even suggest the availability of the port option in the `.ember-cli` file or through the `--port` option.
 
 # Drawbacks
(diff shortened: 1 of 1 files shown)


active/0000-autopick-available-port.md
```diff
@@ -0,0 +1,34 @@
...
+
+Users might not pay attention to the build output by default when they run `ember serve` and might just prefer to see a failure to launch than us handling it.
+
+# Alternatives
+
+Users will have to figure out if another process is using the default port vs their ember app already running and proceed to work around it themselves. 
+
+# Unresolved questions
+
+In live reload's case incrementing by 1 doesn't matter as most users wont care which port live reload works out of. But if port `4200` is not available, what intervals do we use to pick the next available port? This might be important since we'd want the port number to be simple enough to remember.
```
stefanpenner: I suspect if we auto-choose, we should be careful too not risk colliding with other less flexible systems (which may attempt to start up after this server starts). For that, selecting a port in the unassigned range 49152–65535 is most appropriate.

sivakumar-kailasam:
@stefanpenner updated based on your feedback

sivakumar-kailasam:
@stefanpenner if this looks good, I'll get to raising a PR

nathanhammond:
We support this behavior in the most recent beta if you specify `ember s --port 0`. It will automatically choose the port for you and then launch the server. Does that meet your goal, @sivakumar-kailasam?

sivakumar-kailasam:
@nathanhammond Wasn't aware of that. Noticed the relevant discussion in this PR, ember-cli/ember-cli/pull/5391. Is that behavior documented? Would it make sense for it to be added in the [port option's description](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/serve.js#L22)?

stefanpenner:
> is that behavior documented? Would it make sense for it to be added in the port option's description?

yes most likely

sivakumar-kailasam:
Ok will close this in favor of a PR to add it to the description

PR #55, opened 2016-06-15
Anonymous AMD proposal

ef4:
[Rendered](https://github.com/ember-cli/rfcs/blob/anonymous-amd/active/0000-anonymous-amd.md)

Related to https://github.com/ember-cli/ember-cli/pull/5976

### DIFF
--- a/active/0000-anonymous-amd.md
+++ b/active/0000-anonymous-amd.md
@@ -8,7 +8,44 @@ This RFC proposes extending the `app.import` API to consume anonymous AMD module
 
 # Motivation
 
-AMD modules come in two flavors: named and anonymous. _Anonymous AMD_ is the more portable distribution format, but it typically requires preprocessing into _named AMD_ before it can be included into an application. Today, ember-cli users can add arbitrary third-party _named AMD_ modules into their application via:
+AMD modules come in two flavors: named and anonymous. _Anonymous AMD_ is the more portable distribution format, but it typically requires preprocessing into _named AMD_ before it can be included into an application.
+
+    /* Anonymous AMD Examples */
+
+    // direct value
+    define({ color: 'black' });
+
+    // function returning value
+    define(function() { return { color: 'black' }; });
+
+    // function returning value, with declared dependencies
+    define(["jquery", "moment"], function(jQuery, moment) {
+      return {
+        injectTime: function() {
+          jQuery('#time-box').html(moment().format('HH:MM'));
+        }
+      }
+    });
+
+    /* Named AMD Examples */
+
+    // direct value
+    define('my-config', { color: 'black' });
+
+    // function returning value
+    define('my-config', function() { return { color: 'black' }; });
+
+    // function returning value, with declared dependencies
+    define('time-utils', ["jquery", "moment"], function(jQuery, moment) {
+      return {
+        injectTime: function() {
+          jQuery('#time-box').html(moment().format('HH:MM'));
+        }
+      }
+    });
+
+
+Today, ember-cli users can add arbitrary third-party _named AMD_ modules into their application via:
 
     app.import('/path/to/module.js');
 
--- a/active/0000-anonymous-amd.md
+++ b/active/0000-anonymous-amd.md
@@ -59,17 +59,33 @@ In order to de-anonymize AMD, it's necessary to choose a name for the module for
 
 API with an additional argument:
 
-    app.import('/path/to/module.js', { amdModule: 'some-dep' });
+    app.import('/path/to/module.js', {
+      using: [
+        { transformation: 'amd', as: 'some-dep' }
+      ]
+    });
 
-The exact meaning of `amdModule` is: within this Javascript file, any call(s) to the global `define()` function will be intercepted and the given module name (`some-dep` in the above example) will be prepended to the argument list.
+`using` provides a list of transformations. Each transformation is identified by its `transformation` property. Any other properties are treated as arguments to the transformation implementation -- they are opaque to ember-cli. Transformations will run in the given order.
 
-[A complete implementation is available here](https://github.com/ember-cli/ember-cli/pull/5976).
+In this particular case, the `amd` transformation will run and receive the argument `{as: 'some-dep'}`.
+
+The exactly meaning of the `amd` transformation is: within this Javascript file, any call(s) to the global `define()` function will be intercepted and the given module name (`some-dep` in the above example) will be prepended to the argument list.
+
+[A complete implementation is available here](https://github.com/ember-cli/ember-cli/pull/5976). (As of this edit it lags behind updates to this RFC.)
+
+# Learning
+
+An appropriate place to document this feature is [here](https://ember-cli.com/user-guide/#standard-amd-asset). That existing documentation is silent on the distinction between named and anonymous AMD, which probably trips people up.
 
 # Drawbacks
 
 I am not attempting to specify static error detection, mostly because doing that well would require fully parsing and understanding the imported module, which is likely to be more expensive and fragile.
 
 Examples of static errors that would theoretically be nice to detect would be the presence of a _named AMD_ module in the file, the lack of any AMD module in the file, or the present of multiple _anonymous AMD_ modules in the file.
+
+The current implmentation causes any sourcemap information inside the imported file to be discarded (you don't get an invalid sourcemap, but you lose detail).
+
+I have not specified a pluggable way to add additional transformations. My intent is to reserve space in our public API so that future extraction and pluggability is fully backward compatible.
 
 # Alternatives
 
--- a/active/0000-anonymous-amd.md
+++ b/active/0000-anonymous-amd.md
@@ -1,0 +1,45 @@
+- Start Date: (fill me in with today's date, 2016-06-14)
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes extending the `app.import` API to consume anonymous AMD modules. It accompanies [ember-cli PR 5976](https://github.com/ember-cli/ember-cli/pull/5976).
+
+# Motivation
+
+AMD modules come in two flavors: named and anonymous. _Anonymous AMD_ is the more portable distribution format, but it typically requires preprocessing into _named AMD_ before it can be included into an application. Today, ember-cli users can add arbitrary third-party _named AMD_ modules into their application via:
+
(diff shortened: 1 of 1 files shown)


active/0000-anonymous-amd.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Detailed design
+
+In order to de-anonymize AMD, it's necessary to choose a name for the module for use within a given application. So I propose extending the:
+
+    app.import('/path/to/module.js');
+
+API with an additional argument:
+
+    app.import('/path/to/module.js', { amdModule: 'some-dep' });
```
stefanpenner: playing with some variations, thoughts on:
- `{ amd: { id: 'some-dep'}}`
- `{ amd: 'some-dep'}`
- `{ is: 'amd', as: 'some-dep' }`

stefanpenner: its worth noting, we _may_ (although quite likely we wont bother) want to expand this to support bundles of AMD modules, that may have more then 1 export that needs re-mapping. Is there a complementary API that we could use to enable that?

active/0000-anonymous-amd.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Alternatives
+
+Many libraries fall back to global variables if they cannot detect a valid AMD loader. I suspect this is the most common alternate pattern that's in use in the community.
+
+Some applications include their own manually written shims in `vendor` or elsewhere.
+
+# Unresolved questions
+
+We should confirm that my implementation performs well in apps with very large dependency directories.
```
stefanpenner, line 98: i can test the current implementation in one of my very large apps tomorrow

active/0000-anonymous-amd.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Drawbacks
+
+I am not attempting to specify static error detection, mostly because doing that well would require fully parsing and understanding the imported module, which is likely to be more expensive and fragile.
+
+Examples of static errors that would theoretically be nice to detect would be the presence of a _named AMD_ module in the file, the lack of any AMD module in the file, or the present of multiple _anonymous AMD_ modules in the file.
+
+# Alternatives
+
+Many libraries fall back to global variables if they cannot detect a valid AMD loader. I suspect this is the most common alternate pattern that's in use in the community.
```
abuiles, line 92: @ef4 @stefanpenner some libs normally include a check for `define.amd` before calling define (https://github.com/moment/moment/blob/develop/moment.js#L9) but `loader.js` uses `define.petal` https://github.com/ember-cli/loader.js/blob/master/lib/loader/loader.js#L201 .. this will probably need to transform that line too?

ef4, line 92: If I'm understanding you correctly, no.

My implementation just provides a shim around the module that provides its own `define` with `amd`.

active/0000-anonymous-amd.md
ef4: Remapping a bunch of _named_ AMD modules to other names seems like a pretty different (and niche) concern. And multiple _anonymous_ AMD modules in one bundle seems like invalid input.

Of the variants, `{ amd: 'some-dep' }` seems fine, `{ amd: { id: 'some-dep' }}` can be supported cleanly in the future if we ever need to expand the arguments, and `{is: 'amd', as: 'some-dep' }` invites two undefined cases (when either property appears alone).

stefanpenner: > Remapping a bunch of named AMD modules to other names seems like a pretty different (and niche) concern. And multiple anonymous AMD modules in one bundle seems like invalid input.

Ya, exists for legacy reasons. Merely sharing thoughts re: API naming.

active/0000-anonymous-amd.md
```diff
@@ -0,0 +1,45 @@
...
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes extending the `app.import` API to consume anonymous AMD modules. It accompanies [ember-cli PR 5976](https://github.com/ember-cli/ember-cli/pull/5976).
+
+# Motivation
+
+AMD modules come in two flavors: named and anonymous. _Anonymous AMD_ is the more portable distribution format, but it typically requires preprocessing into _named AMD_ before it can be included into an application. Today, ember-cli users can add arbitrary third-party _named AMD_ modules into their application via:
```
Turbo87: > AMD modules come in two flavors: named and anonymous

could you maybe include examples or links to examples here?

krisselden:
This is awesome

active/0000-anonymous-amd.md
nathanhammond, line 98: Updates on this?

nathanhammond:
I'm in favor of this change. Other approaches to do things like this include running your own processing step and then shoving the result into the `vendor` tree. This dramatically improves that story.

My only hesitations are small:
- It's likely that this may make it harder to support some build time tooling for rollup-like behavior as the implementation has the files in a separate tree.
- Perf cost of already-slow `define` and wrapping it in another IIFEs (though @krisselden can speak to that far better than I can; I'm cargo-culting based upon conversations overheard).

Note that I'm in favor even with those hesitations.

active/0000-anonymous-amd.md
ef4: Good idea, I added some examples to the document.

ef4:
@nathanhammond can you elaborate on "may make it harder to support some build time tooling for rollup-like behavior"? I don't fully understand that part.

Perhaps you're concerned that my implementation is hiding the real module inside an extra layer? The API doesn't wed us to that behavior -- I think the API leaves us free to alter the implementation.

stefanpenner:
@ef4 this looks great, three things:
- learning section, how do we teach this? (basically how to make it easy to find/here about)
- this AMD isn't really a full tree of AMD rather a single file. We should be sure to document this.
- current implementation breaks source-maps (might be fine, our current story here isn't great yet anyways)

active/0000-anonymous-amd.md
stefanpenner, line 98: testing now.

stefanpenner, line 98: my massive app:

I thought this introduced a perf hit, but then i noticed `master` has a regression (i will need to investigate) cold boots jumped 25% and rebuilds doubled, will investigate. But the PR associated with this seems largely fine.

before: 

```
cold boot: 95856ms.
warm boot: 16441ms.
rebuild: 2210ms
```

after (without any AMD modules)

```
warm boot: 15626ms
rebuild: 2262ms.
```

after (with any AMD modules)

```
warm boot: 15441ms
rebuild: 2223ms
```

Conclusion, no obvious issues introduced by this feature in my app (others may, especially if they app.import a massive amount of things, but i suspect we are ok...)

ef4, line 98: My code short-circuits if you have no AMD imports configured. It would be nice to confirm that there's no massive hit when you turn on the first one.

stefanpenner, line 98: @ef4 i miss typed above, the third example is with 1 amdModule.

stefanpenner, line 98: fixed my perf regression (unrelated to this RFC/PR) no change basically \w amdModule being actually used.

```
rebuild (\w amdModule) 720ms.
```

stefanpenner:
@ef4 friendly ping

ef4:
@stefanpenner I pushed some edits that address your questions, except for

> this AMD isn't really a full tree of AMD rather a single file. We should be sure to document this.

AFAIK, `app.import` already doesn't do trees, so that is not specific to this issue. Plus I really don't know what it would mean to import a tree of anonymous AMD -- I don't think that's a well-defined thing (how could the modules in the tree find each other if they're really still anonymous?).

I also changed the proposed API based on discussion in https://github.com/ember-cli/ember-cli/pull/5976 so that we reserve API-space for future transformations, which can eventually become pluggable.

active/0000-anonymous-amd.md
```diff
@@ -0,0 +1,98 @@
...
+# Detailed design
+
+In order to de-anonymize AMD, it's necessary to choose a name for the module for use within a given application. So I propose extending the:
+
+    app.import('/path/to/module.js');
+
+API with an additional argument:
+
+    app.import('/path/to/module.js', {
+      using: [
```
stefanpenner, line 63: good pick, I like this.

stefanpenner:
@ef4 this has been in PFC for some time, did the core team reach a consensus that I missed or did we drop the ball?

If so, I do feel good about this change. Should we maybe give core until this friday  to object, then we ship?

stefanpenner:
this has survived PFC

active/0000-anonymous-amd.md
```diff
@@ -0,0 +1,98 @@
...
+        { transformation: 'amd', as: 'some-dep' }
+      ]
+    });
+
+`using` provides a list of transformations. Each transformation is identified by its `transformation` property. Any other properties are treated as arguments to the transformation implementation -- they are opaque to ember-cli. Transformations will run in the given order.
```
nathanhammond, line 68: We didn't actually implement this API in our current implementation. The loop that we run here iterates over the transforms, errors if it isn't AMD, and then sets the last one:
https://github.com/ember-cli/ember-cli/blob/e9a29f8246340a60b585f75825b2a64ae9001b72/lib/broccoli/ember-app.js#L1563-L1577

```javascript
app.import('vendor/anonymous-amd-example.js', {
  using: [
    { transformation: 'amd',  as: 'hello-world'},
    { transformation: 'amd',  as: 'this-is-dumb-but-demonstrates-the-bug'}
  ],
  outputFile: '/assets/output.js'
});
```

We then process the resulting pre-clobbered `amdModuleNames` property without regards for order of any additional transformations, or even aware of the transformations array at all:
https://github.com/ember-cli/ember-cli/blob/e9a29f8246340a60b585f75825b2a64ae9001b72/lib/broccoli/ember-app.js#L1005-L1014

/cc @ef4 @stefanpenner

stefanpenner: Ya, API in RFC don't always equate to the final outcome. Largely they exist to make sure we are on the right track, but changes can happen during implementation.

I think though, we may still not have documented stuff.

RobbieTheWagner: So is there a way we could support the multiple transforms? I'm having a hard time figuring out how to use fastboot-transform with the amd transform. Would love to see support for multiple happening in a row 😄

PR #59, opened 2016-06-30
auto-blueprint-template-include

kellyselden:
[Rendered](https://github.com/kellyselden/ember-cli-rfcs/blob/auto-blueprint-template-include/active/0000-auto-blueprint-template-include.md)

### DIFF
--- a/active/0000-auto-blueprint-template-include.md
+++ b/active/0000-auto-blueprint-template-include.md
@@ -1,0 +1,54 @@
+- Start Date: 2016-06-30
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+Treat template blueprints like test blueprints, they are their own blueprint and
+included automatically.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+I want to override the blueprint for either a route or a component. I want to
+make changes to the js file only and not worry about copying the default
+template from upstream along with the default replace value logic. Vice-versa
+for modifying templates without worrying about the js file.
+
+# Detailed design
+
+We create blueprints for `component-template` and `route-template`. We move the
+template blueprints to these folders. These blueprints are automatically called
+like the test blueprints because they end in `-template`.
+
+We then either deprecate the `template` blueprint because it is confusing
+whether it is a route or component template, or we just leave it. Or we can just
+delegate that decision to the legacy blueprint project or ember addon.
+
+Making this a bit more concrete:
+
+MOVE FROM: https://github.com/ember-cli/ember-cli-legacy-blueprints/blob/master/blueprints/component/files/__root__/__templatepath__/__templatename__.hbs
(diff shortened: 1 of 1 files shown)


nathanhammond:
Making this a bit more concrete, the proposal is to:

MOVE FROM: https://github.com/ember-cli/ember-cli-legacy-blueprints/blob/master/blueprints/component/files/__root__/__templatepath__/__templatename__.hbs
TO: https://github.com/ember-cli/ember-cli-legacy-blueprints/blob/master/blueprints/component-template/files/__root__/__templatepath__/__templatename__.hbs

MOVE FROM: https://github.com/ember-cli/ember-cli-legacy-blueprints/blob/master/blueprints/route/files/__root__/__templatepath__/__templatename__.hbs
TO: https://github.com/ember-cli/ember-cli-legacy-blueprints/blob/master/blueprints/route-template/files/__root__/__templatepath__/__templatename__.hbs

DEPRECATE: https://github.com/ember-cli/ember-cli-legacy-blueprints/tree/master/blueprints/template

And the underlying changes that support that automatic invocation.

nathanhammond:
@kellyselden What particular experience precipitated you opening this RFC?

This RFC allows a user to override the template without also having to reinclude the upstream JS, making it more composable. I'd call this better-factored, but rather than doing piece-wise fixes to this I'd be interested in seeing if we can come up with the "bigger picture" solution to making it easier to compose chains of blueprints. If we do this we would be making a change for this behavior and then making a change again whenever we get around to figuring out how to make blueprints play nicely together.

If those are going to be far enough apart on the timeline I don't care as it won't feel like churn. If we're aiming to do that grand rethinking soon I'd hold off on this. (Minimum one year.) We might can skirt around my churn concerns on this with changing semantics to support the module unification story.

/cc @kimroen

kellyselden:
@nathanhammond The experience was overriding the default route blueprint, and unexpectedly having to copy over the default template blueprint to retain that functionality.

trabus:
I really like this idea, and it would be relatively easy to implement (aside from the busy work of updating the tests). I do think we should leave the original `template` blueprint, as there is still a need to generate templates outside of the auto-generation. Furthermore, I think we can get away with extending (now that we're using core-object) the `template` blueprint for the `route-template` and `component-template` blueprints, so we won't have to maintain three different blueprints across multiple projects (ember, ember-cli-legacy-blueprints, etc).

nathanhammond:
Per team meeting we need to:
- Include the concrete information from [this comment](https://github.com/ember-cli/rfcs/pull/59#issuecomment-230080911) into the RFC itself.
- Explore alternative solution of expanding the `template` blueprint.
- Check impact of things like `partial` on the module unification story and blueprint generation.

kellyselden:
@nathanhammond updated per your above first two comments.

> Check impact of things like partial on the module unification story and blueprint generation.

Can you explain this more? I don't see a `partial` blueprint.

Turbo87:
closing due to inactivity

PR #61, opened 2016-07-24
add RFC: JSON file blueprint merging

rtablada:
[rendered](https://github.com/rtablada/rfcs/blob/json-file-blueprint-merge/active/0000-json-file-blueprint-merge.md)

### DIFF
--- a/active/0000-json-file-blueprint-merge.md
+++ b/active/0000-json-file-blueprint-merge.md
@@ -1,0 +1,33 @@
+- Start Date: 2016-07-24
+- RFC PR:
+- ember-cli Issue:
+
+# Summary
+
+This feature would allow blueprints and ember upgrades to interactively merge JSON files instead of wiping away existing files, or causing syntax errors due to manual text editing.
+
+# Motivation
+
+The current process for including things like `bower.json`, `package.json`, `.babelrc`, etc leads the user to have to manually read through diff logs and resolve change issues.
+Most often, this occurs with the `ember-cli` app upgrade blueprint.
+Especially for newer developers, it can be confusing why some of your addons get uninstalled in the process of upgrading.
+
+# Detailed design
+
+Since both the existing and suggested JSON files can be parsed and read as JSON, it makes sense that we could diff each key and value programmatically and have more information than a simple text comparison.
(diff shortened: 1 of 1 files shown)


rwjblue:
I mentioned some of this in slack, but I figured it makes sense to detail it here also.  I came up with a "hair brained" idea to just handle the dep/devDep changes in `package.json` and `bower.json` a little while back. Then I successfully nerd sniped @bantic, who implemented it in https://github.com/bantic/package-merger.  

The idea is to take the three input sources and emit the exact changes that are needed (roughly similar to what fs-tree-diff does).  [This test file](https://github.com/bantic/package-merger/blob/master/tests/merge.js) is a great walk through of what it does.  Using this information, we can do the right thing, without requiring user interaction.

I'd like to see this incorporated into the RFC...

bantic:
The package-merger util mentioned above could be used to help users automatically upgrade their (dev)dependencies — if you feed it the source (current ember-cli blueprint'd package.json) and next (target version of ember-cli's blueprint'd package.json) it can compare them and spit out a data structure showing which deps need to be added, which deps should be removed, and which just need to be changed/updgraded. It has the nice feature that packages that the user explicitly added won't be removed by the ember-cli upgrade, and packages that they explicitly removed (e.g., if they remove ember-data because they don't use it) will be ignored rather than re-added. This has been a pain point for me in the past — the diff view of package.json wants to remove all the dependencies that aren't part of the stock ember-cli install.

Perhaps there is a fallback to generic json key-value diffing for json files that aren't recognized, and for ones that we know are for managing deps (bower.json and package.json), the 3-way merge package-merger tool can be used?

nathanhammond:
Notes from today's core team meeting:
- Current implementation reorders the keys, this is actually a meaningful thing inside of package.json.
- Let's adopt the `package-merger` work from @bantic and use that as the basis for this feature.

@rtablada @bantic Can the two of y'all work together to update this RFC and get into a state where we can move it to final comment period?

bantic:
Yeah, would be happy to! Will follow up in community slack.

Turbo87:
closing due to inactivity

PR #62, opened 2016-07-30
Extract Default CSS Handling

nathanhammond:
Extract the built-in CSS handling inside of Ember CLI to an addon (`ember-cli-css`) which registers itself identically to the way `ember-cli-sass` or `ember-cli-less` functions.

[Rendered.](https://github.com/nathanhammond/ember-cli-rfcs/blob/extract-default-css-handling/active/0000-extract-default-css-handling.md)

### DIFF
--- a/active/0000-extract-default-css-handling.md
+++ b/active/0000-extract-default-css-handling.md
@@ -1,0 +1,47 @@
+- Start Date: 2016-07-30
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+Extract the built-in CSS handling inside of Ember CLI to an addon (`ember-cli-css`) which registers itself identically to the way `ember-cli-sass` or `ember-cli-less` functions.
+
+# Motivation
+
+The default handling of CSS inside of Ember CLI is ad hoc and generally underspecified. It functions as a primitive CSS preprocessor but doesn't register itself as such. There are a lot of issues that we experience as a consequence of this approach:
+- [We literally have questions in the documentation.](https://github.com/ember-cli/ember-cli.github.io/blame/6205e1f73da9a4edceeb5be70a93ef058b223975/_posts/2013-04-09-asset-compilation.md#L130)
+- The outcome of any particular CSS setup is left to the imagination and reverse-engineering of the user–approximately programming by permutation.
+- There is no way to configure what happens inside of addons without manually writing a `treeForStyle` hook.
+- [Behavior of underlying packages](https://github.com/ember-cli/ember-cli/pull/5874) can break builds which can force us to release new versions of Ember CLI.
+- The behavior of CSS handling is unnecessarily tied to the Ember CLI version and it makes the process of landing improvements to our CSS handling much more difficult than truly necessary.
+
+Moving it to an addon allows us to:
+- Iterate faster.
+- Not couple the behavior to Ember CLI.
+- Provide understandable guarantees about what happens inside of addons when they include `ember-cli-css`.
+- Make it more clear that this is something which can be swapped out for something like `ember-cli-sass`.
+
+# Detailed Design
+
+The first step in the process is to simply extract all of the default CSS handling out into an addon which implements the exact same behavior as we presently have. For the first release there will be _no_ changes in the outcome of the build. This will become part of the default Ember CLI blueprint. This allows us to meet our backwards compatibility guarantees.
+
+After the extraction we will then be able to allow our users to select which version of the `ember-cli-css` addon they wish to use which allows us to start moving it toward reasonable defaults for application- and addon-level behavior.
+
+# How We Teach This
+
+The primary change will be to our documentation at [ember-cli.com](https://ember-cli.com/user-guide/#css). We'll need to make it more clear that you must swap the default-included `ember-cli-css` behavior for any of the other preprocessors, and that the default CSS version functions identically to a preprocessor.
+
+Further we should use this opportunity to add additional documentation about the inclusion of preprocessors inside of addons and how that functionality works.
+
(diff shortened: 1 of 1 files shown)


nathanhammond:
Some documentation I just wrote describing the insanity that is our CSS processing model.

---

CSS files will be built in the same way that they are inside of typical adddons. What does that mean? I'm glad you asked.
1. All nested addons are processed. Each of them may return a `style` tree. By default these style trees only contain the contents of `addon/styles/addon.css`. The contents of the `addon/styles/addon.css` file is moved inside of the Broccoli tree to `${addon-name}.css`.
2. All top-level addons (those directly depended upon by the host) have all of `addon/styles/**/*.css` included into the host's `vendor.css` file. For example `addon/styles/foo.css` will appear in the output Broccoli tree at `foo.css`.
3. If you name a CSS file in one of the top-level addons the same as an addon name (e.g. addon name is `alpha`), any top-level addon which has a CSS file of the same name as that addon (`alpha.css`) and is provided by an addon lexically after it (`zeta`) will clobber the contents of
   `alpha/addon/styles/addon.css` (from anywhere in the dependency graph) with `zeta/addon/styles/alpha.css`.

active/0000-extract-default-css-handling.md
```diff
@@ -0,0 +1,47 @@
...
+- Iterate faster.
+- Not couple the behavior to Ember CLI.
+- Provide understandable guarantees about what happens inside of addons when they include `ember-cli-css`.
+- Make it more clear that this is something which can be swapped out for something like `ember-cli-sass`.
+
+# Detailed Design
+
+The first step in the process is to simply extract all of the default CSS handling out into an addon which implements the exact same behavior as we presently have. For the first release there will be _no_ changes in the outcome of the build. This will become part of the default Ember CLI blueprint. This allows us to meet our backwards compatibility guarantees.
+
+After the extraction we will then be able to allow our users to select which version of the `ember-cli-css` addon they wish to use which allows us to start moving it toward reasonable defaults for application- and addon-level behavior.
```
MiguelMadero, line 28: Different addons with different versions of ember-cli-css may lead to a lot of confusion or incompatible configurations. This is probably out of scope for the rfc and we should deal with this concerns once we start iterating on it.

active/0000-extract-default-css-handling.md
```diff
@@ -0,0 +1,47 @@
...
+
+# How We Teach This
+
+The primary change will be to our documentation at [ember-cli.com](https://ember-cli.com/user-guide/#css). We'll need to make it more clear that you must swap the default-included `ember-cli-css` behavior for any of the other preprocessors, and that the default CSS version functions identically to a preprocessor.
+
+Further we should use this opportunity to add additional documentation about the inclusion of preprocessors inside of addons and how that functionality works.
+
+# Drawbacks
+
+When users first upgrade to an Ember CLI version that doesn't include the CSS behavior by default they will be _required_ to add the addon to their `package.json`. As most people use more advanced preprocessors this is likely to be a small audience.
```
cibernox, line 38: Since, as you mention, most people uses css pre-processors, it might worth adding a flag to the CLI to generate apps with those preprocessors already, at least with sass, since it's by far the most popular one.
- `ember new my-app --sass`
- `ember new my-app --less`

It would default to `--css` and install this new addon.
That also means that it must be a community-blessed sass & less addons, but in practice, at least for sass, there is already one.

Gaurav0:
Do we have adequate test coverage on how CSS is built today to ensure that we can actually extract code to ember-cli-css in a backward compatible way?

How the heck is an addon going to place its CSS in such a way as to be compatible with different versions of ember-cli-css? How is it going to test this in an automated fashion?

active/0000-extract-default-css-handling.md
nathanhammond, line 28: This is no different from our current preprocessor story and has, in practice, not been an issue. Since all of this work happens in Node-land it should hopefully "just work" and resolve to the correct version of the preprocessor.

I'm not proposing any changes to how CSS works in Ember CLI at this point and I believe that's something that we address inside of the addon as it matures.

active/0000-extract-default-css-handling.md
nathanhammond, line 38: I'm hesitant to add a larger API surface area–especially for "blessed but not maintained by us" addons. [Our current documentation](https://ember-cli.com/user-guide/#stylesheets) provides the one-liner `ember install ember-cli-sass` as the means to adopt a different solution. I believe that is Good Enough™?

nathanhammond:
@Gaurav0 Test coverage would clearly be ensured.

As for compatibility of addons with multiple versions of `ember-cli-css` I suspect that nobody has undertaken the research I did to identify what the actual behavior is and are therefore not relying on it. At this point the only guarantee addons have is that their `addon.css` file will appear somewhere inside of the `vendor.css` file. Depending upon where they are in the dependency graph they could also have other CSS they include present in the build output, but that's unreliable by virtue of their not controlling at which level they are included.

We should move to deprecate the [behavior I describe here](https://github.com/ember-cli/rfcs/pull/62#issuecomment-236400248) and limit the result to just the `addon.css` file and then it becomes easy to ensure compatibility.

Further, the addon itself may include `ember-cli-css` to handle its preprocessing according to the version attached to that addon.

active/0000-extract-default-css-handling.md
stefanpenner, line 38: Yes, i think customizing the default blueprint is a different story. Although this RFC would be a step in the right direction.

cibernox, line 38: Out of the table then. I also think that things are easy enough today to not really worry about this.

rwjblue:

Thank you!

PR #63, opened 2016-08-06
Nested addons

xcambar:
This RFC follows the discussion (and PR) from https://github.com/ember-cli/ember-cli/pull/6111

[rendered](https://github.com/xcambar/rfcs-1/blob/259693c2cd61b025080c6049e586cb012d8414ad/active/0000-nested-addon-imports.md)

### DIFF
--- a/active/0000-nested-addon-imports.md
+++ b/active/0000-nested-addon-imports.md
@@ -1,0 +1,63 @@
+- Start Date: (fill me in with today's date, 2016-08-06)
+- RFC PR: 
+- ember-cli Issue: 
+
+# Summary
+
+The new asset import API, available since ember-cli 2.7.0, offers the possibility to 
+import Ember CLI addons from Ember CLI addons (aka, nested addons).
+This RFC intends to find solutions for typical edge cases.
+
+# Motivation
+
+Since ember-cli 2.7.0, addons can be nested if the nesting floows the following
+scenario:
+* The nested addons do not make use of the property `parent.bowerProperty` to define
+  assets paths because it not always defined (it is not defined in EmberAddon instances)
+* The nested addons do not need to execute a default blueprint.
+
+Solving these use cases would allow for a simpler, more generic, addons development
+process, and will allow teams to build more reusable and higher level addons for
+their projects and the community (like [ember-admin](https://github.com/DockYard/ember-admin)
+using [ember-power-select](ember-power-select.com) with no additional boilerplate, for instance).
+
+# Detailed design
+
+Here is a rundown of the questions raised with @nathanhammond in [a previous PR](https://github.com/ember-cli/ember-cli/pull/6111):
+
+1. Should nested addons be allowed to use their grandparents' `bower_components` directory?
+1. How/When should the default blueprints of the nested be run?
+1. How to deal with conflicting versions (The app uses my-addon 1.0.0, but my-addon 2.0.0 is required as a nested component)?
+
+For each of these questions, here's a first attempt towards an answer:
+
+1. Yes, they should, because there's only one bower directory available to the app, and it should be available to the whole app,
+  including to the dependencies. See [the PR linked above](https://github.com/ember-cli/ember-cli/pull/6111) for an example.
+1. They could be run in the "afterInstall" hook, when the dependencies (hence nested addons) are already installed.
+  See [this comment](https://github.com/ember-cli/ember-cli/pull/6111#issuecomment-235829717) for a specific implementation
+1. Ideally, this falls under Bower's (or npm's) umbrella.
+
+The same way that ember-cli hides the complexity for installing addons and the related node deps and bower deps,
+the initial idea is to hide the complexity to do (most of) the hard work in ember-cli itself.
+
+# Drawbacks
+
+We should be extra careful with this, as we are likely to fall in one of the
+many traps that package managers have to face, dealing with conflicts, resolutions, ordering, tree buiding, etc...
+The more we can rely on the package managers, the less code to embark in ember-cli, the better.
+
+# Alternatives
+
+Instead of hiding the complexity and have ember-cli orchestrate the installation of the complex addons by itself,
+exposing new APIs to let the developer do as they must for the nested addons would let us see the use cases and find
+opportunities to simplify later.
+
+Following that, the alternative could be to:
+* make the `EmberApp#_findHost` API public, and proably rename it. (answers the `bowerDirectory` problem with: `this.findHost().bowerDirectory`)
+* expose in the EmberAddon instance the addons it depends from.
+* expose a new API to execute blueprints (answers the blueprint execution problem with the above)
+* Let Bower handle conflicting versions (answers the versions' conflict resolution)
+
+# Unresolved questions
+
+Many. The document will be updated as they emerge.


active/0000-nested-addon-imports.md
```diff
@@ -0,0 +1,63 @@
...
+Here is a rundown of the questions raised with @nathanhammond in [a previous PR](https://github.com/ember-cli/ember-cli/pull/6111):
+
+1. Should nested addons be allowed to use their grandparents' `bower_components` directory?
+1. How/When should the default blueprints of the nested be run?
+1. How to deal with conflicting versions (The app uses my-addon 1.0.0, but my-addon 2.0.0 is required as a nested component)?
+
+For each of these questions, here's a first attempt towards an answer:
+
+1. Yes, they should, because there's only one bower directory available to the app, and it should be available to the whole app,
+  including to the dependencies. See [the PR linked above](https://github.com/ember-cli/ember-cli/pull/6111) for an example.
```
nathanhammond, line 35: Having spent more time thinking about this, I'm opposed to this for reasons of version conflicts. Addons could attempt to `bower install` something which is in direct conflict with the host applications' dependencies. Further we'd be encouraging further adoption of `bower` when our intent is to move farther away from that.

In general if we do this I believe that the addon itself should receive its own separate `bower_components` directory.

This in general creates the situation where the host needs to be aware of the addon's dependencies in order to not conflict, which I think breaks the degree of isolation that we should be aiming for.

active/0000-nested-addon-imports.md
```diff
@@ -0,0 +1,63 @@
...
+1. They could be run in the "afterInstall" hook, when the dependencies (hence nested addons) are already installed.
+  See [this comment](https://github.com/ember-cli/ember-cli/pull/6111#issuecomment-235829717) for a specific implementation
```
nathanhammond, line 37: We install a lot of addons as `devDependencies` which will make this especially exciting for users (lots of things which simply won't be present on a default `npm install` for the host application. We would need to actually teach people how this all works and write documentation to make it clear.

active/0000-nested-addon-imports.md
```diff
@@ -0,0 +1,63 @@
...
+1. Ideally, this falls under Bower's (or npm's) umbrella.
```
nathanhammond, line 38: Right now we've got a solution where the "last one wins" as it simply does a `broccoli-merge` of the directories. Figuring out which one is last is actually an `undefined` implementation detail, however. It's lexicographically sorted, but also depth sorted. How we should support this remains unclear to me.

I think we'd need a module isolation per addon ability to make this work and rewrite `import`s in the addon to point to a particular version if we're going to support Node-style module scoping instead of a single universal module scope.

active/0000-nested-addon-imports.md
```diff
@@ -0,0 +1,63 @@
...
+
+We should be extra careful with this, as we are likely to fall in one of the
+many traps that package managers have to face, dealing with conflicts, resolutions, ordering, tree buiding, etc...
+The more we can rely on the package managers, the less code to embark in ember-cli, the better.
+
+# Alternatives
+
+Instead of hiding the complexity and have ember-cli orchestrate the installation of the complex addons by itself,
+exposing new APIs to let the developer do as they must for the nested addons would let us see the use cases and find
+opportunities to simplify later.
```
nathanhammond, line 53: Later: I believe that this is the right time to be discussing this topic in advance of implementing https://github.com/emberjs/rfcs/pull/143.

active/0000-nested-addon-imports.md
```diff
@@ -0,0 +1,63 @@
...
+
+Following that, the alternative could be to:
+* make the `EmberApp#_findHost` API public, and proably rename it. (answers the `bowerDirectory` problem with: `this.findHost().bowerDirectory`)
```
nathanhammond, line 56: `_findHost` implements something that a _lot_ of addons have already been doing. 😛 This is a reasonable proposal, but something that we should probably wait to address until we finish Lazy Loading Engines. (I abused this API when writing that build.)

active/0000-nested-addon-imports.md
```diff
@@ -0,0 +1,63 @@
...
+* expose in the EmberAddon instance the addons it depends from.
```
nathanhammond, line 57: Can you rephrase? I'm not sure I understand what you mean by this.

active/0000-nested-addon-imports.md
```diff
@@ -0,0 +1,63 @@
...
+* expose a new API to execute blueprints (answers the blueprint execution problem with the above)
+* Let Bower handle conflicting versions (answers the versions' conflict resolution)
```
nathanhammond, line 59: But those conflicts would be preserved in the host application, which makes addons leak information into their hosts and me sad.

active/0000-nested-addon-imports.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Summary
+
+The new asset import API, available since ember-cli 2.7.0, offers the possibility to 
+import Ember CLI addons from Ember CLI addons (aka, nested addons).
+This RFC intends to find solutions for typical edge cases.
+
+# Motivation
+
+Since ember-cli 2.7.0, addons can be nested if the nesting floows the following
```
nathanhammond, line 13: Nested addons has been possible for a _long_ time. `this.import` just makes it easier to do the right thing.

active/0000-nested-addon-imports.md
```diff
@@ -0,0 +1,63 @@
...
+scenario:
+* The nested addons do not make use of the property `parent.bowerProperty` to define
+  assets paths because it not always defined (it is not defined in EmberAddon instances)
+* The nested addons do not need to execute a default blueprint.
```
nathanhammond, line 17: `ember install` is what executes that blueprint, so we'd need to abstract that into something that can be called on `npm install` of the overall project.

stefanpenner:
Improving nested add-ons is important, I have two points to raise.
- we plan to drop bower, only provided it as legacy. As such I would prefer we avoid "fixing" bower in nested dependencies.
- how does this plan deal with conflicting dependencies and dupes?

active/0000-nested-addon-imports.md
xcambar, line 35: My concern with your proposal is that by implementing it, we will build a better/safer Bower in ember-cli on top of Bower, which, as you made clear, we want to move away from.

Looking towards what's "best", I agree with your description, but IMHO, it is an order of magnitude more complicated than warning the user that "though Bower is supported, it appears that installing AddonX requires the installation of dependencyA#1.0.0 which conflicts with the currently installed dependencyA#2.0.0".

A message does not solve anything but at the same time it does not cost much, and maybe that's enough for  a part of the stack that will eventually be deprecated.

active/0000-nested-addon-imports.md
xcambar, line 57: Add a method/property to `EmberAddon` so that it is easy/possible to list their children addons.

The goal is to ease the execution of the necessary operations on the children addons to complete the installation process.

active/0000-nested-addon-imports.md
xcambar, line 13: I confess I've been "a bit generous" with 2.7.0 😆

xcambar:
@stefanpenner 👍 on not fixing Bower.

About dupes and conflicts, I would like, as much as possible, to rely on npm for that.
As I would love to drop Bower, I would let the user know that a conflict _may_ happen (see https://github.com/ember-cli/rfcs/pull/63#discussion_r74641811).

rwjblue:

Thank you!

PR #67, opened 2016-09-12
macro blueprint

kellyselden:
[Rendered](https://github.com/kellyselden/ember-cli-rfcs/blob/macro-blueprint/active/0000-macro-blueprint.md)

### DIFF
--- a/active/0000-macro-blueprint.md
+++ b/active/0000-macro-blueprint.md
@@ -34,6 +34,39 @@ import myMacro from 'my-project/macros/my-macro';
 
 myValue: myMacro('key1', 'key2')
 ```
+We would also supply the accompanying unit test blueprint:
+
+``` js
+import <%= camelizedModuleName %> from '<%= dasherizedModulePrefix %>/macros/<%= dasherizedModuleName %>';
+import { module, test } from 'qunit';
+
+const MyType = Ember.Object.extend({
+  computedProperty: <%= camelizedModuleName %>(/* keys... */)
+});
+
+let myObj;
+
+module('<%= friendlyTestName %>', {
+  beforeEach() {
+    myObj = MyType.create({
+      // add any source data to feed your computed macro
+      // key1: 'value1'
+      // ...
+    });
+  }
+});
+
+// Replace this with your real tests.
+test('it works', function(assert) {
+  // you can also alter any source data per test
+  // myObj.set('key1', 'value2');
+  // ...
+  let result = myObj.get('computedProperty');
+  assert.ok(result);
+});
+```
+(I'm a little verbose in test setup and comments because I believe our current test blueprints are not verbose enough.)
+
 # Drawbacks
 
 The code to structure a computed macro is so simple, that it may be unnecessary to turn it into a blueprint. Although I would argue that beginners may not know how simple it is. It may seem like a daunting task to start making your own macros. This could help start them off.
--- a/active/0000-macro-blueprint.md
+++ b/active/0000-macro-blueprint.md
@@ -1,0 +1,47 @@
+# macro blueprint RFC
+
+- Start Date: 2016-09-12
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+Add a `macro` blueprint to the default list of blueprints.
+
+# Motivation
+
+To make it slightly easier for people to make custom computed macros. The syntax of a custom macro is already dead simple, but having a default structure, starting point, and folder location would help beginners get started.
+
+# Detailed design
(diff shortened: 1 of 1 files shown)


active/0000-macro-blueprint.md
```diff
@@ -0,0 +1,47 @@
...
+``` js
+import myMacro from 'my-project/macros/my-macro';
+
+//...
+
+myValue: myMacro('key1', 'key2')
+```
+# Drawbacks
+
+The code to structure a computed macro is so simple, that it may be unnecessary to turn it into a blueprint. Although I would argue that beginners may not know how simple it is. It may seem like a daunting task to start making your own macros. This could help start them off.
```
hjdivad, line 72: Fair enough, although one could say the same thing about many blueprints.  Even if the blueprints do little more than expose a convention, there's value in that.

active/0000-macro-blueprint.md
```diff
@@ -0,0 +1,47 @@
...
+export default function(/* keys... */) {
+  return Ember.computed(/* keys... */, function() {
+    // your normal computed code
+  });
+}
+```
+Then once filled in, could be used like:
+
+``` js
+import myMacro from 'my-project/macros/my-macro';
```
nathanhammond, line 31: This works in terms of module unification, I just dislike squatting on another top-level collection.

nathanhammond:
The existing patterns around this are all set by [`ember-cpm`](https://github.com/cibernox/ember-cpm/blob/master/blueprints/macro/files/addon/__path__/__name__.js)

I'm not sure if we get a ton of benefit moving this into Ember directly? Is it enough to point people to ember-cpm?

/cc @cibernox

cibernox:
I personally think that an entry in the docs is enough. Besides, the blueprint would just generate

``` js
export default function(/*...keys*/) {
  return Ember.computed(/*...keys*/, function() {

  });
}
```

I think that the `util` blueprint mostly does this already, and the user has to understand how to computed properties work to create one anyway.

kellyselden:
> I think that the util blueprint mostly does this already, and the user has to understand how to computed properties work to create one anyway.

The `util` blueprint doesn't help in this case, because it gives you no guidance in how to structure the macro. This blueprint is meant to bridge the gap in helping a beginner understand how to set up a custom computed macro. I remember when I was starting out, I assumed the macros were another one of the deep magical things that ember does for you, and good luck if you want to customize.

I think having a blueprint as a starting point and a `macros/` folder "convention" would really help out newbies.

acorncom:
> I think having a blueprint as a starting point and a macros/ folder "convention" would really help out newbies

I would concur.  I've been around / using Ember for 2+ yrs now and am just learning that this is even an option 😀

active/0000-macro-blueprint.md
cibernox: I also think that macros are a particular type of util. 

@nathanhammond Would `app/utils/macros/my-macro` play nicely with the module-unification?

kellyselden:
Some things that came out of the cli core team meeting:
- This is a good thing, but needs ember core team approval.
- Including an accompanying test blueprint would help solidify the usefulness
- Should it live in `app/utils/macros` instead of `app/macros`?

kellyselden:
I've added a testing blueprint to the RFC.

locks:
I believe it should live at `app/macros` like other official collections, as well as some of the community provided solutions. If there happens to be a name collision it should be possible to address on a case by case.
On the why not to use `app/utils/macros`, `app/utils` is the junk drawer of Ember applications. Suddenly doing nested collections, and doing it in `app/utils` seems confusing.

tomdale:
While computed property macros are very powerful and useful, I'm not sure we can afford to add more conceptual weight to the Ember getting started experience. My preference is to keep the core constrained to the bare minimum amount of concepts, and leave education about macros (and thus this generator) to popular community addons like `ember-cpm`.

nathanhammond:
@tomdale Does that comment represent the general consensus of the core team per today's meeting?

cibernox:
I also subscribe that idea of narrow core, but I think that CP macros are useful enough to deserve an entry in the guides about how to create your own, maybe just after this point: https://guides.emberjs.com/v2.8.0/object-model/computed-properties/#toc_computed-property-macros

tomdale:
@nathanhammond I believe so.

nathanhammond:
@locks et al.: This then falls into the area documentation of advanced patterns in the Ember community and addons that we have which gracefully enable them. Just drawing this to your attention as another reference point.

Closing in favor of addon, not default-included in the `app` or `addon` blueprint.

PR #68, opened 2016-09-16
Support blueprint generation without accompanying test

SaladFork:
[Rendered](https://github.com/SaladFork/ember-cli-rfcs/blob/blueprint-without-test/active/0000-blueprint-without-test.md)


kellyselden:
It should be noted that this has been discussed at length on other issues, with the general consensus that having it a "little" painful to not do testing is a good thing.

Is the added complexity in our app worth the effort of discarding files before you check in your code?

I do recognize that there are a lot of people that don't want to test and just want ember-cli to get out of their way.

Balancing act all around...

nathanhammond:
@SaladFork I'm one of the major proponents (if not the biggest proponent) of making it _slightly_ more awkward to not have tests. The other thing is that in your example I also expect that at _some_ point your "simple" model will morph into something more complicated. When it morphs into something more complicated having an already-generated test file is nice because it makes it clear where the tests are supposed to go.

All-in-all the users who would ask for this behavior seem to be like you: people who know their way around the library incredibly well and comfortable digging themselves out of any hole they find themselves in. The easier it is to do something that's not a good practice the more likely people who have less familiarity will use it.

In short: API design is a feature, including the addition of friction into the API in select places. I'm therefore opposed to this. We'll discuss at a future Ember CLI weekly meeting and circle back.

kellyselden:
> We'll discuss at a future Ember CLI weekly meeting and circle back.

And as always, anyone is welcome to attend!

Turbo87:
closing due to inactivity

PR #71, opened 2016-10-10
rfc to add liveReloadFilters property

bantic:
[rendered](https://github.com/bantic/rfcs/blob/e76c5876ed2118119e68769a2dcd5dd8b84f80c9/active/0000-add-livereload-filters-property.md)

refs https://github.com/ember-cli/ember-cli/issues/6192

### DIFF
--- a/active/0000-add-livereload-filters-property.md
+++ b/active/0000-add-livereload-filters-property.md
@@ -1,0 +1,90 @@
+- Start Date: 2016-10-10
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+This RFC suggests exposing a project property called `liveReloadFilters`. This is an array of
+strings or regexes that will be used to filter *out* paths from ember-cli's array of post-build changed filepaths that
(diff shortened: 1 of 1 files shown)


briangonzalez:
@bantic Has any work begun on this one?

bantic:
@briangonzalez Nope, no work done yet. I was hoping to get some more comments here and get a sense whether this would be ratified or needs more thought. Looks like you found a use case for this as well in https://github.com/dollarshaveclub/ember-cli/commit/a01de486a518276972596e534785cb80b7f1cc0a. Any thoughts on the RFC above?

briangonzalez:
@bantic I read the whole thing over and I like the approach of keeping the old config around as to not break folks' current implementation.

Have you been able to take a peek at the code from https://github.com/dollarshaveclub/ember-cli/commit/a01de486a518276972596e534785cb80b7f1cc0a. We have it working for us locally, but I'd love others to give it a shot, too.

lukemelia:
I could use this today. Your proposed solution seems reasonable to me @bantic. 👍

rwjblue:

Thank you!

PR #74, opened 2016-10-20
Nested configuration rfc

pzuraq:
[rendered](https://github.com/pzuraq/rfcs/blob/nested-configuration-rfc/active/0000-nested-configuration.md)

### DIFF
--- a/active/0000-nested-configuration.md
+++ b/active/0000-nested-configuration.md
@@ -1,0 +1,610 @@
+- Start Date: 2016-10-05
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+Provides a unified system for configuring nested engines and addons from their
+hosts. This system aims to be flexible enough to enable different configurations
+per app/engine, while still being simple and manageable at scale.
+
+# Motivation
+Currently, the method for configuring an Application or Engine (a Host) is
+directly through modules from the `/config/environment.js` file. Each Host only
+has access to its own environment file, and there is no way for a Host to
+configure its children. This means that in order for an Engine to have the same
+configuration as its Host, all of the environment properties must be manually
+duplicated from the parent into the engine. As applications grow in size and more
+parts get broken out into engines, this inflexibility can make apps harder to maintain.
+
+Furthermore, there is no equivalent for Addons as their environment file
+doesn't get compiled into the final application. There are some community-based
+solutions which inject the root application's environment config into the addon,
+but this results in tight coupling and should be avoided if possible.
+
+Design constraints and considerations for the unified system:
+
+1. It should statically compile all configurations at build time. This will help
+to prevent strange bugs due to changing configurations, and will result in
+better overall performance.
+
+1. It should continue to allow engines to have their own configuration.
+
+1. It should allow engines to specify an external API of overrideable
+configuration variables.
+
+1. It should allow hosts to specify different configurations for engines based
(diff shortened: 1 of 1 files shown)


knownasilya:
Surprising nobody has commented on this. Figure it'll get more attention once engines are more main stream.

stefanpenner:
> Surprising nobody has commented on this.

largely time being a limiting function :(

pzuraq:
Definitely understandable. I'd still like to land this when there's more time to discuss the API, it's something that would really help addon developers.

stefanpenner:
@pzuraq we should make this a priority to review/work through post holiday.

lukemelia:
@pzuraq I came across this RFC when considering a particular scenario. I'd like to describe it, along with my understanding of how it would be addressed in your proposal.

I have an app "dashboard" that depends on the addon "ember-freestyle" to provide tools for a styleguide. It also depends on the addon "yapp-ember-kit", which has it's own dependency on the addon "ember-freestyle". So the dependency tree looks like:

```
yapp-dashboard (app)
    |--ember-freestyle (addon)
    |--yapp-shared (addon)
       |--ember-freestyle (addon)
```

I would like to exclude the "ember-freestyle" addon from my production builds, in both instances. Note that I control the "yapp-dashboard" app and the "yapp-shared" addon, but not the "ember-freestyle" addon.

Based on what you proposed, if ember-freestyle had an "enabled" option, I would have something like this:

```
// yapp-dashboard/config/addons/ember-freestyle.js
export default {
  enabled: false  
}
```

How would I make this conditional, though, so that I would disable this only when `environment === "production"`?

If ember-freestyle did not have an `enabled` flag, then I would want to use ember-cli's [addon blacklisting feature](https://github.com/ember-cli/ember-cli/pull/5107). ember-cli doesn't yet support this feature for nested addons, AFAIK. How would you see that fitting into this proposed syntax?

pzuraq:
@lukemelia So this is what the `getConfig` helper and function parameters were supposed to address, in tandem. To make it conditional, you would do something like this:

```js
// yapp-dashboard/config/addons/ember-freestyle.js
export default {
  enabled() {
    // `yapp-dashboard.environment` would refer to the object exported by 
    // `yapp-dashboard/config/environment.js`, so there would have to be 
    // another property on that object to actually contain the environment string.
    return getConfig('yapp-dashboard.environment.environment') === 'production';
  }
}
```

The functions would be executed at _build_ time and their output would replace the property on the final object. Also, you could do this configuration within `yapp-shared` to share it among apps, but `yapp-dashboard`'s configuration would end up winning in the end as it is the host.

Looking back on this, the `getConfig` helper is a little clumsy. It has some constraints that are not immediately apparent (you can only `getConfig` of a value higher up in the tree), and it's behavior is a little magical/ill-defined. I think it could definitely use some work, it may make sense to pass the previous level of configuration directly into the next one (like how `config/environment` receives the environment value.)

pzuraq:
Also, thinking about how you would be able to do this from `yapp-shared`, it may make sense to use a generic placeholder to refer to the current host's config rather than the app name:

```js
// yapp-shared/config/addons/ember-freestyle.js
export default {
  enabled() {
    return getConfig('host.environment.environment') === 'production';
  }
}
```

rwjblue:

Thank you!

pzuraq:
For sure, this is super out of date, definitely think it could still be useful but I also think we need to rethink it 👍

PR #77, opened 2016-11-11
RFC to build fastboot in ember-cli.

kratiahuja:
[rendered](https://github.com/kratiahuja/rfcs/blob/add-fastboot/active/0000-build-fastboot-in-ember-cli.md)

Corresponding PR which does the work described in this RFC is [here](https://github.com/ember-cli/ember-cli/pull/6395).

Long term RFC which should allow this to become a generic feature is described in this [RFC](https://github.com/ember-cli/rfcs/pull/75)

cc: @stefanpenner @tomdale @rwjblue @ryanone @masonwan @kristoferbaxter @arjansingh

### DIFF
--- a/active/0000-build-fastboot-in-ember-cli.md
+++ b/active/0000-build-fastboot-in-ember-cli.md
@@ -4,27 +4,27 @@
 
 # Summary
 
-This RFC will primarily allow fastboot assets to be built using `ember build`. Fastboot is a different environment in which your ember apps run and require your app behavior to change a bit in this environment that running your app in the browser. This requires building fastboot assets that compliment the current browser assets.
+This RFC will primarily allow FastBoot assets to be built using `ember build`. FastBoot is a different environment in which your ember apps run and require your app behavior to change a bit in this environment that running your app in the browser. This requires building FastBoot assets that compliment the current browser assets.
 
-This RFC describes the details of how fastboot will be built inside `ember-cli`. Since we do not know the generic usecases, we would like to build this `ember-cli` as the first pass. Once we have understood the usecases, we can make it generic per this [RFC](https://github.com/ember-cli/rfcs/pull/75). This RFC aims to expose the minimal public API and work to build fastboot in `ember-cli`.
+This RFC describes the details of how FastBoot will be built inside `ember-cli`. Since we do not know the generic usecases, we would like to build this into `ember-cli` as the first pass. Once we have understood the usecases, we can make it generic per this [RFC](https://github.com/ember-cli/rfcs/pull/75). This RFC aims to expose the minimal public API and work to build FastBoot in `ember-cli`.
 
 # Motivation
-[Fastboot](https://ember-fastboot.com) allows ember apps to be rendered on the server side. It renders the initial content of your app allowing your app to be used in SEO usecases. On a very high level, fastboot loads your assets in a Node environment and creates the Ember App once the assets are loaded. It does not do any automatic routing but creates Ember application instance whenever there is a server side rendering request and runs your route to render the initial render in Node.
+[FastBoot](https://ember-FastBoot.com) allows ember apps to be rendered on the server side. It renders the initial content of your app allowing your app to be used in SEO usecases. On a very high level, FastBoot loads your assets in a Node environment and creates the Ember App once the assets are loaded. It does not do any automatic routing but creates `Ember.ApplicationInstance` whenever there is a server side rendering request and runs your route to render the initial render in Node.
 
-In order for your app/addon to be able to run in a Node environment, it needs to be compatible in such environment. Compatibility includes not doing DOM/window access in Node, making sure you are able to make API requests in Node etc. Therefore an app/addon may need to provide fastboot specific initializers, instance-initializers, services etc that are only applicable in the Node environment and not in the browser. Including them in the browser has performance implication since you will send down more bytes to the client.
+In order for your app/addon to be able to run in a Node environment, it needs to be compatible in such an environment. Compatibility includes not doing DOM/window access in Node, making sure you are able to make API requests in Node etc. Therefore an app/addon may need to provide FastBoot specific initializers, instance-initializers, services etc that are only applicable in the Node environment and not in the browser. Including them in the browser has performance implication since you could be sending down more bytes than required to the client at runtime.
 
-Currently fastboot builds the app twice and generates a different set of assets for fastboot and browser environments via ember fastboot command. Most of the app between the two sets of assets is same except where we want fastboot to override the browser behavior or need fastboot specific handling. Currently ember fastboot filters these and generates the two sets of assets. Building different sets of assets (app and vendor files) for different environments cannot guarantee the app works correctly in both environments always. Moreover, ember-fastboot --serve-assets command also spins up a node server and serves your assets. This is very similar to what ember serve does today. However, we end up forking and creating two parallel worlds with problems such as serving stale assets, creating race conditions during serving and building etc. We want to be able converge the two worlds such that fastboot should be able to specify the last middleware that will be used to serve the assets in fastboot. This helps us not create fastboot commands and keeps the eco-system unified.
+Currently FastBoot builds the app twice and generates a different set of assets for FastBoot and browser environments via `ember fastboot` command. Most of the app between the two sets of assets is same except where we want FastBoot to override the browser behavior or need FastBoot specific handling. Currently `ember fastboot` filters these and generates the two sets of assets (each for a different environment). Due to filtering and creating two different sets of assets we end up running the build on the app twice. Building different sets of assets (app and vendor files) for different environments cannot guarantee the app works correctly in both environments always. This RFC helps us not create FastBoot commands (which patches the private API of `ember-cli`, does a double build to generate two sets of assets and does a heavy filtering in the app tree causing large build times) for build and keeps the eco-system unified.
 
-We want to build these fastboot specific code as an additive asset to the browser that loads in the Node VM in addtion to the the other assets. If it overrides any browser behavior, having the same module id will override the definition in the loader. In order to build fastboot assets in a performant way, we would like to integrate building fastboot into ember-cli. This issue specifically tracks integrating building fastboot in ember-cli. When we need to build more such artifacts, we can extend this to be a generic target solution.
+We want to build this FastBoot specific code as an additive asset to the browser that loads in the Node VM in addition to the the other assets. If it overrides any browser behavior, having the same module id will override the definition in the loader. In order to build FastBoot assets in a performant way, we would like to integrate building FastBoot into ember-cli. This issue specifically tracks integrating building FastBoot in ember-cli. When we need to build more such artifacts, we can extend this to be a generic target solution.
 
 # Detailed design
 
-## Directory Structure for fastboot app/addons
-In order to build the fastboot assets which are an addition to the app and vendor files, we want to define a separate directory structure. This will allow fastboot specific app code to live in its own namespace and does not touch the browser specific app file.
+## Directory Structure for FastBoot app/addons
+In order to build the FastBoot assets which are an addition to the app and vendor files, we want to define a separate directory structure. This will allow FastBoot specific app code to live in its own namespace and does not touch the browser specific app file.
 
-Currently today no fastboot specific code lives in the addon namespace. Hence we do not need `fastboot/addon` namespace as part of this RFC in `ember-cli`.
+*Note*: Currently today no FastBoot specific code lives in the addon namespace. Hence we do not need `FastBoot/addon` namespace and building of that namespace as part of this RFC.
 
-For example an addon containing fastboot specific functionality will have a directory structure like:
+For example an addon containing FastBoot specific functionality will have a directory structure like:
 ```
 my-addon/
 |-- addon/
@@ -41,30 +41,34 @@ my-addon/
 ```
 Note: The same applies for an app as well.
 
-This allows us to keep fastboot specific functionality in its own namespace and allows us to build fastboot artifact by walking only the subset of the tree rather than walking the whole app and doing filtering (as done right now).
+This allows us to keep FastBoot specific functionality in its own namespace and allows us to build FastBoot artifact by walking only the subset of the tree rather than walking the whole app and doing filtering (as done right now).
 
 After an ember build, the generated assets will be:
 1. vendor.js
 2. app.js
 3. app-fastboot.js
 
-In the Node VM, we will first load the assets and eval them in the above order. Therefore, if you wanted to override a functionality in fastboot, you can keep the name of the file same in fastboot namespace and will override it when app-fastboot.js gets loaded and evaluated. This also guarantees that we do not regress existing ember builds that do not contain the fastboot namespace.
+In the Node VM, we will first load the assets and eval them in the above order. Therefore, if you wanted to override a functionality in FastBoot, you can keep the name of the file same in FastBoot namespace and will override it when app-fastboot.js gets loaded and evaluated. This also guarantees that we do not regress existing ember builds that do not contain the FastBoot namespace.
 
 ## Public API
-With building the additional assets, this RFC would also like to expose some public APIs so that addons that are doing fastboot specific things can leverage that to build fastboot assets. The additional public API will be as follows:
-1. `preprocessTree(type, tree)`: Currently the type includes only for browser specific assets. We want to invoke this hook for fastboot asset as well. Therefore, when fastboot assets are built, this hook will be invoked with type as `fastboot`. Addons can use this hook to munge the fastboot/app tree before it gets processed.
-2. `postprocessTree(type, tree)`: This currently is also only invoked for browser specific assets. We want to invoke the same hook with `fastboot` as the type when building fastboot asset.
-3. `treeForFastbootApp(tree)`: This is a new hook that this RFC proposes to add which can be used by addons to munge the tree for `fastboot/app` before it is processed.
+With building the additional assets, this RFC would also like to expose some public APIs so that addons that are doing FastBoot specific things can leverage that to build FastBoot assets. The additional public API will be as follows:
+1. `preprocessTree(type, tree)`: Currently the type includes only for browser specific assets. We want to invoke this hook for FastBoot asset as well. Therefore, when FastBoot assets are built, this hook will be invoked with type as `app-fastboot`. Addons can use this hook to munge the FastBoot/app tree before it gets processed.
+2. `postprocessTree(type, tree)`: This currently is also only invoked for browser specific assets. We want to invoke the same hook with `app-fastboot` as the type when building FastBoot asset.
+3. `treeForFastBootApp(tree)`: This is a new hook that this RFC proposes to add which can be used by addons to munge the tree for `fastboot/app` before it is processed.
 
 # How We Teach This
 
 This is an advanced feature and any app that defines `fastboot/app` and runs `ember build` will end up generating an additional asset as `app-fastboot.js`. We would need to update the `ember-cli` documentation to explain users how this additional asset is generated and how it will be of limited use.
 
-Moreover, `ember-cli-fastboot` will do the work of moving the existing fastboot specific funtionality in various addons into `fastboot/app`. As part of building fastboot assets faster, we will deprecate the `ember fastboot` command.
+Moreover, `ember-cli-fastboot` will do the work of moving the existing FastBoot specific funtionality in various addons into `fastBoot/app`. As part of building FastBoot assets faster, we will deprecate the `ember fastBoot` command.
+
+We will also be upgrading [FastBoot](https://ember-fastboot.com) website of how this is works with `ember-cli`.
 
 # Drawbacks
 
-The only drawback this has is that it builds fastboot specific code in `ember-cli`. This is only being done since we do not want to come up with a generic solution until we fully know the usecases to make this generic that any additional asset can be built.
+Following are the drawbacks of this approach:
+- [ ] It builds FastBoot specific code in `ember-cli`. This is only being done since we do not want to come up with a generic solution until we fully know the usecases to make this generic that any additional asset can be built.
+- [ ] This approach of loading additive assets after the main asset heavily depends on the loader.js behavior of "last one" wins always. If this behavior changes in future, this approach will break.
 
 # Alternatives
 
@@ -72,4 +76,4 @@ An alternative [RFC](https://github.com/ember-cli/rfcs/pull/75) which is a more 
 
 # Unresolved questions
 
-N/A
+- [ ] How will this work in long term with module unification?
--- a/active/0000-build-fastboot-in-ember-cli.md
+++ b/active/0000-build-fastboot-in-ember-cli.md
@@ -1,0 +1,75 @@
+- Start Date: 2016-11-10
+- RFC PR: (leave this empty)
+- Ember CLI Issue: (leave this empty)
+
+# Summary
+
+This RFC will primarily allow fastboot assets to be built using `ember build`. Fastboot is a different environment in which your ember apps run and require your app behavior to change a bit in this environment that running your app in the browser. This requires building fastboot assets that compliment the current browser assets.
+
+This RFC describes the details of how fastboot will be built inside `ember-cli`. Since we do not know the generic usecases, we would like to build this `ember-cli` as the first pass. Once we have understood the usecases, we can make it generic per this [RFC](https://github.com/ember-cli/rfcs/pull/75). This RFC aims to expose the minimal public API and work to build fastboot in `ember-cli`.
+
+# Motivation
+[Fastboot](https://ember-fastboot.com) allows ember apps to be rendered on the server side. It renders the initial content of your app allowing your app to be used in SEO usecases. On a very high level, fastboot loads your assets in a Node environment and creates the Ember App once the assets are loaded. It does not do any automatic routing but creates Ember application instance whenever there is a server side rendering request and runs your route to render the initial render in Node.
+
+In order for your app/addon to be able to run in a Node environment, it needs to be compatible in such environment. Compatibility includes not doing DOM/window access in Node, making sure you are able to make API requests in Node etc. Therefore an app/addon may need to provide fastboot specific initializers, instance-initializers, services etc that are only applicable in the Node environment and not in the browser. Including them in the browser has performance implication since you will send down more bytes to the client.
+
+Currently fastboot builds the app twice and generates a different set of assets for fastboot and browser environments via ember fastboot command. Most of the app between the two sets of assets is same except where we want fastboot to override the browser behavior or need fastboot specific handling. Currently ember fastboot filters these and generates the two sets of assets. Building different sets of assets (app and vendor files) for different environments cannot guarantee the app works correctly in both environments always. Moreover, ember-fastboot --serve-assets command also spins up a node server and serves your assets. This is very similar to what ember serve does today. However, we end up forking and creating two parallel worlds with problems such as serving stale assets, creating race conditions during serving and building etc. We want to be able converge the two worlds such that fastboot should be able to specify the last middleware that will be used to serve the assets in fastboot. This helps us not create fastboot commands and keeps the eco-system unified.
+
+We want to build these fastboot specific code as an additive asset to the browser that loads in the Node VM in addtion to the the other assets. If it overrides any browser behavior, having the same module id will override the definition in the loader. In order to build fastboot assets in a performant way, we would like to integrate building fastboot into ember-cli. This issue specifically tracks integrating building fastboot in ember-cli. When we need to build more such artifacts, we can extend this to be a generic target solution.
+
+# Detailed design
+
+## Directory Structure for fastboot app/addons
+In order to build the fastboot assets which are an addition to the app and vendor files, we want to define a separate directory structure. This will allow fastboot specific app code to live in its own namespace and does not touch the browser specific app file.
+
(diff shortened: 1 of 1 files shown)


kratiahuja:
Some history are mentioned in the following issues:
1. [ember-cli issue](https://github.com/ember-cli/ember-cli/issues/6350)
2. [ember-cli-fastboot issue](https://github.com/ember-fastboot/ember-cli-fastboot/issues/264)

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,75 @@
+- Start Date: 2016-11-10
+- RFC PR: (leave this empty)
+- Ember CLI Issue: (leave this empty)
+
+# Summary
+
+This RFC will primarily allow fastboot assets to be built using `ember build`. Fastboot is a different environment in which your ember apps run and require your app behavior to change a bit in this environment that running your app in the browser. This requires building fastboot assets that compliment the current browser assets.
+
+This RFC describes the details of how fastboot will be built inside `ember-cli`. Since we do not know the generic usecases, we would like to build this `ember-cli` as the first pass. Once we have understood the usecases, we can make it generic per this [RFC](https://github.com/ember-cli/rfcs/pull/75). This RFC aims to expose the minimal public API and work to build fastboot in `ember-cli`.
```
rwjblue: > would like to build this `ember-cli` as the first pass

Sounds a bit odd, maybe you mean:

> would like to build this into `ember-cli` as the first pass

?

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Motivation
+[Fastboot](https://ember-fastboot.com) allows ember apps to be rendered on the server side. It renders the initial content of your app allowing your app to be used in SEO usecases. On a very high level, fastboot loads your assets in a Node environment and creates the Ember App once the assets are loaded. It does not do any automatic routing but creates Ember application instance whenever there is a server side rendering request and runs your route to render the initial render in Node.
```
rwjblue: >  It does not do any automatic routing but creates Ember application instance whenever there is a server side rendering request and runs your route to render the initial render in Node.

Slight tweak for correctness:

>  It does not do any automatic routing but creates an `Ember.ApplicationInstance` instance whenever there is a server side rendering request and runs your applications initial render in Node.

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,75 @@
...
+
+In order for your app/addon to be able to run in a Node environment, it needs to be compatible in such environment. Compatibility includes not doing DOM/window access in Node, making sure you are able to make API requests in Node etc. Therefore an app/addon may need to provide fastboot specific initializers, instance-initializers, services etc that are only applicable in the Node environment and not in the browser. Including them in the browser has performance implication since you will send down more bytes to the client.
```
rwjblue: >  Including them in the browser has performance implication since you will send down more bytes to the client.

Might read better as:

>  Including them in the browser has performance implications since you would be sending down more bytes than required to the client at runtime.

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,75 @@
...
+
+Currently fastboot builds the app twice and generates a different set of assets for fastboot and browser environments via ember fastboot command. Most of the app between the two sets of assets is same except where we want fastboot to override the browser behavior or need fastboot specific handling. Currently ember fastboot filters these and generates the two sets of assets. Building different sets of assets (app and vendor files) for different environments cannot guarantee the app works correctly in both environments always. Moreover, ember-fastboot --serve-assets command also spins up a node server and serves your assets. This is very similar to what ember serve does today. However, we end up forking and creating two parallel worlds with problems such as serving stale assets, creating race conditions during serving and building etc. We want to be able converge the two worlds such that fastboot should be able to specify the last middleware that will be used to serve the assets in fastboot. This helps us not create fastboot commands and keeps the eco-system unified.
```
rwjblue: > ember-fastboot --serve-assets command 

Should be:

> `ember fastboot --serve-assets` command

rwjblue: > However, we end up forking and creating two parallel worlds with problems such as serving stale assets, creating race conditions during serving and building etc.

Should also mention 2X build speed issues

rwjblue: > We want to be able converge the two worlds such that fastboot should be able to specify the last middleware that will be used to serve the assets in fastboot.

This is slightly confusing to me (from the RFC's perspective "what are middlewares?", "why do I care about the last middleware?", etc), can you tweak a bit to make it clearer?

rwjblue: > This helps us not create fastboot commands and keeps the eco-system unified.

We should explain why we care about not duplicating commands and whatnot...

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,75 @@
...
+
+We want to build these fastboot specific code as an additive asset to the browser that loads in the Node VM in addtion to the the other assets. If it overrides any browser behavior, having the same module id will override the definition in the loader. In order to build fastboot assets in a performant way, we would like to integrate building fastboot into ember-cli. This issue specifically tracks integrating building fastboot in ember-cli. When we need to build more such artifacts, we can extend this to be a generic target solution.
```
rwjblue: > We want to build these fastboot specific code as an additive asset to the browser

s/these/this/

rwjblue: > If it overrides any browser behavior, having the same module id will override the definition in the loader.

We should discuss (somewhere, possibly in unanswered questions) what this means if/when we use ember-cli apps without a loader (i.e. if we switch to using rollup and treeshaking to avoid needing to transpile to AMD).  I'd hate to have gone forward with this RFC and close the door on other large important build related refactors.

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Detailed design
+
+## Directory Structure for fastboot app/addons
+In order to build the fastboot assets which are an addition to the app and vendor files, we want to define a separate directory structure. This will allow fastboot specific app code to live in its own namespace and does not touch the browser specific app file.
+
+Currently today no fastboot specific code lives in the addon namespace. Hence we do not need `fastboot/addon` namespace as part of this RFC in `ember-cli`.
```
rwjblue: > Currently today no fastboot specific code lives in the addon namespace.

This isn't strictly true.  Addons can and do use fastboot specific things (generally by using the `fastboot-filter-initializers` package).  See [ember-cli-head](https://github.com/ronco/ember-cli-head) as an example.

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,75 @@
...
+After an ember build, the generated assets will be:
+1. vendor.js
+2. app.js
+3. app-fastboot.js
+
+In the Node VM, we will first load the assets and eval them in the above order. Therefore, if you wanted to override a functionality in fastboot, you can keep the name of the file same in fastboot namespace and will override it when app-fastboot.js gets loaded and evaluated. This also guarantees that we do not regress existing ember builds that do not contain the fastboot namespace.
+
+## Public API
+With building the additional assets, this RFC would also like to expose some public APIs so that addons that are doing fastboot specific things can leverage that to build fastboot assets. The additional public API will be as follows:
+1. `preprocessTree(type, tree)`: Currently the type includes only for browser specific assets. We want to invoke this hook for fastboot asset as well. Therefore, when fastboot assets are built, this hook will be invoked with type as `fastboot`. Addons can use this hook to munge the fastboot/app tree before it gets processed.
```
rwjblue: > this hook will be invoked with type as `fastboot`

This should be `fastboot-app` and `fastboot-addon`

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,75 @@
...
+2. `postprocessTree(type, tree)`: This currently is also only invoked for browser specific assets. We want to invoke the same hook with `fastboot` as the type when building fastboot asset.
```
rwjblue: Same as above, should be `app-fastboot` and `addon-fastboot`

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,75 @@
...
+3. `treeForFastbootApp(tree)`: This is a new hook that this RFC proposes to add which can be used by addons to munge the tree for `fastboot/app` before it is processed.
```
rwjblue: Should also include a `treeForFastbootAddon` hook

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,75 @@
...
+
+# How We Teach This
+
+This is an advanced feature and any app that defines `fastboot/app` and runs `ember build` will end up generating an additional asset as `app-fastboot.js`. We would need to update the `ember-cli` documentation to explain users how this additional asset is generated and how it will be of limited use.
```
rwjblue, line 61: Should include guides / docs in ember-fastboot.com

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,75 @@
...
+
+Moreover, `ember-cli-fastboot` will do the work of moving the existing fastboot specific funtionality in various addons into `fastboot/app`. As part of building fastboot assets faster, we will deprecate the `ember fastboot` command.
+
+# Drawbacks
+
+The only drawback this has is that it builds fastboot specific code in `ember-cli`. This is only being done since we do not want to come up with a generic solution until we fully know the usecases to make this generic that any additional asset can be built.
```
rwjblue: Relies on `loader.js` behavior of "last wins", which is not guaranteed to be true for:
- a world where we don't use `loader.js` (i.e. if we move to rollup built assets)
- lazy engines?

rwjblue (commented):
Notes:
- Various references to fastboot should be "FastBoot".
- Is there a specific reason to rely on the "last module registered wins" clobbering behavior as opposed to relying on broccoli-merge-trees (with `{ overwrite: true }`) to allow only the final concat to differ?
- I would have thought that we would want a `browser/app` and `browser/addon`. This way the main generic app bundles live in `addon` or `app` and the "platform specific" parts live in either `browser/` or `fastboot/`.  The resulting build would have `app.js`, `browser-app.js`, `fastboot-app.js`, and possibly a `combined-app.js` (which would be the main app bundle along with the main browser bundle).

active/0000-build-fastboot-in-ember-cli.md
ryanone: `fastboot-app` (as stated above), or `app-fastboot`?   It seems that `fastboot-app` is more appropriate given how the mapping currently looks:

```
{
  app: 'treeForApp',
  styles: 'treeForStyles',
  templates: 'treeForTemplates',
  'addon-templates': 'treeForAddonTemplates',
  'fastboot-app': 'treeForFastbootApp' // ...?
}
```

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: @rwjblue If I understand correctly today:
1. No fastboot specific code lives in addon name space (including `ember-cli-head`
2. `fastboot-filter-initializers` broccoli plugin only filter out instance-initializers and initializers and I don't see why someone would have those in addon namespace.

In general the goal of this RFC was:
1. To unblock Fastboot 1.0
2. Do the minimal work in ember-cli such that fastboot can be built. 

While I agree we can have fastboot specific things in addon namespace (there is nothing prevents that today), there is no usecase I have seen that is happening currently. I really want to address what is needed to unblock Fastboot 1.0. Having, addon namespace for fastboot is _good to have_ feature but not a _required_ feature.

I can add a long term section of allowing building `fastboot/addon` if you want.

kratiahuja:
> I would have thought that we would want a browser/app and browser/addon. This way the main generic app bundles live in addon or app and the "platform specific" parts live in either browser/ or fastboot/. The resulting build would have app.js, browser-app.js, fastboot-app.js, and possibly a combined-app.js (which would be the main app bundle along with the main browser bundle).

I agree about this as a "long term" goal (see this [RFC](https://github.com/ember-cli/rfcs/pull/77)). If we want to do this with this RFC, it then it means this is a backward incompatible change in `ember-cli`. The goal of this RFC was:
1. Keep `ember-cli` backward compatible
2. Do not regress the current `ember-cli` builds that builds browser specific assets.
3.  Do the minimal things in `ember-cli` such that FastBoot 1.0 is unblocked. 

If everyone agrees to the above scope of the problem, this this is what this RFC is supposed to address.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: @rwjblue I do not want to add building `fastboot/addon` since there is no usecase today. In future, if we have a usecase I would agree we want to expose that hook. To satisfy the long term needs (which are undetermined right now), I can add this and about building `fastboot/addon` but we will not be addressing it in the corresponding PR. Thoughts?

kratiahuja:
> Is there a specific reason to rely on the "last module registered wins" clobbering behavior as opposed to relying on broccoli-merge-trees (with { overwrite: true }) to allow only the final concat to differ?

No specific reason, that is how loader.js works today. We can do the final merge of `app.js` and `app-fastboot.js` but this is a behavior I would leave upto the platform to determine. (In case of fastboot, `ember-cli-fastboot` can do that in the `postProcessTree` for `fastboot` type).

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: @rwjblue `lazy-engines` are loaded upfront in FastBoot environment. A world where `loader.js` does not live as well, the eval and ordering of loading the assets in the sandbox will still do the override I believe.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: @ryanone  I think the confusion is from two aspects:
1. The generated fastboot asset is called `app-fastboot.js`
2. The hook is call `fastboot`. 

The reason Rob is suggesting to keep the names same across both the same.

active/0000-build-fastboot-in-ember-cli.md
rwjblue: Hmm, the vendor asset that we load in fastboot should still have `loader.js`, and it is used (AFAIK we do not transpile with other things like rollup in ember-cli-fastboot today so we have to still be relying on loader.js there...).

active/0000-build-fastboot-in-ember-cli.md
rwjblue: Sounds good to me. We can make a more focused RFC for specifically adding this at a later date.

kratiahuja: Sounds good. I am going to drop this then.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: Yes, there is nothing special done in fastboot. The loader definition and API is exposed in the sandbox. 

I was answering the question about the rollup. I honestly don't know what is the plan for that.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: Fixed.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: Fixed.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: Fixed.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: Fixed.

kratiahuja: Fixed explaining the double builds. Apologies the middleware shouldn't have been part of this RFC. That will likely be a different RFC (hopefully coming next week...)

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: Fixed and added that as an open question.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: We discussed this. `fastboot/addon` will not be built as part of this RFC.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: Fixed to use `app-fastboot` and dropped the `fastboot-addon` per the scope of this RFC.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: We discussed this. `fastboot/addon` will not be built as part of this RFC.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: Fixed.

kratiahuja:
@rwjblue updated per your feedback.

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,79 @@
+- Start Date: 2016-11-10
+- RFC PR: (leave this empty)
+- Ember CLI Issue: (leave this empty)
+
+# Summary
+
+This RFC will primarily allow FastBoot assets to be built using `ember build`. FastBoot is a different environment in which your ember apps run and require your app behavior to change a bit in this environment that running your app in the browser. This requires building FastBoot assets that compliment the current browser assets.
```
nathanhammond, line 7: `s/compliment/complement/`

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,79 @@
...
+
+We want to build this FastBoot specific code as an additive asset to the browser that loads in the Node VM in addition to the the other assets. If it overrides any browser behavior, having the same module id will override the definition in the loader. In order to build FastBoot assets in a performant way, we would like to integrate building FastBoot into ember-cli. This issue specifically tracks integrating building FastBoot in ember-cli. When we need to build more such artifacts, we can extend this to be a generic target solution.
+
+# Detailed design
+
+## Directory Structure for FastBoot app/addons
+In order to build the FastBoot assets which are an addition to the app and vendor files, we want to define a separate directory structure. This will allow FastBoot specific app code to live in its own namespace and does not touch the browser specific app file.
+
+*Note*: Currently today no FastBoot specific code lives in the addon namespace. Hence we do not need `FastBoot/addon` namespace and building of that namespace as part of this RFC.
```
nathanhammond, line 25: Should we have a concept of a FatBoot-specific addon? How should this get merged from a child addon which needs to do something different in FastBoot?

I feel like we're missing a concept here that I'm sloppily calling `treeForFastboot()`.

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,79 @@
...
+After an ember build, the generated assets will be:
+1. vendor.js
+2. app.js
+3. app-fastboot.js
+
+In the Node VM, we will first load the assets and eval them in the above order. Therefore, if you wanted to override a functionality in FastBoot, you can keep the name of the file same in FastBoot namespace and will override it when app-fastboot.js gets loaded and evaluated. This also guarantees that we do not regress existing ember builds that do not contain the FastBoot namespace.
+
+## Public API
+With building the additional assets, this RFC would also like to expose some public APIs so that addons that are doing FastBoot specific things can leverage that to build FastBoot assets. The additional public API will be as follows:
+1. `preprocessTree(type, tree)`: Currently the type includes only for browser specific assets. We want to invoke this hook for FastBoot asset as well. Therefore, when FastBoot assets are built, this hook will be invoked with type as `app-fastboot`. Addons can use this hook to munge the FastBoot/app tree before it gets processed.
```
nathanhammond, line 55: Newline before.

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,79 @@
...
+|   |   |   |-- ajax.js
+```
+Note: The same applies for an app as well.
+
+This allows us to keep FastBoot specific functionality in its own namespace and allows us to build FastBoot artifact by walking only the subset of the tree rather than walking the whole app and doing filtering (as done right now).
+
+After an ember build, the generated assets will be:
+1. vendor.js
+2. app.js
+3. app-fastboot.js
```
nathanhammond, line 49: I'm opposed to adding an additional asset and new failure mode here. I'd much rather generate `app-fastboot.js` which is `mergeTrees` + `overwrite: true` here and contains the actual result..

active/0000-build-fastboot-in-ember-cli.md
nathanhammond, line 55: I'm concerned about this being the place that we want to insert things. `preprocessTree` and `postprocessTree` receive broad categories of file types (`test`, `template`, `js`, `css`), not segments of built files. Segmentation tends to be more in the `treeFor` space inside of Ember CLI.

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Public API
+With building the additional assets, this RFC would also like to expose some public APIs so that addons that are doing FastBoot specific things can leverage that to build FastBoot assets. The additional public API will be as follows:
+1. `preprocessTree(type, tree)`: Currently the type includes only for browser specific assets. We want to invoke this hook for FastBoot asset as well. Therefore, when FastBoot assets are built, this hook will be invoked with type as `app-fastboot`. Addons can use this hook to munge the FastBoot/app tree before it gets processed.
+2. `postprocessTree(type, tree)`: This currently is also only invoked for browser specific assets. We want to invoke the same hook with `app-fastboot` as the type when building FastBoot asset.
+3. `treeForFastBootApp(tree)`: This is a new hook that this RFC proposes to add which can be used by addons to munge the tree for `fastboot/app` before it is processed.
```
nathanhammond, line 57: How are you expecting this to be invoked at nested addon levels? What order guarantees are you expecting? And since you're expecting a `tree` as input, what is that `tree`?

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,79 @@
...
+
+Following are the drawbacks of this approach:
+- [ ] It builds FastBoot specific code in `ember-cli`. This is only being done since we do not want to come up with a generic solution until we fully know the usecases to make this generic that any additional asset can be built.
+- [ ] This approach of loading additive assets after the main asset heavily depends on the loader.js behavior of "last one" wins always. If this behavior changes in future, this approach will break.
+
+# Alternatives
+
+An alternative [RFC](https://github.com/ember-cli/rfcs/pull/75) which is a more generic solution and long term goal in `ember-cli`.
+
+# Unresolved questions
```
nathanhammond, line 77: FastBoot needs to be aware of the engine route map as well in order to properly inline assets in response to the initial request.

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,79 @@
...
+
+Moreover, `ember-cli-fastboot` will do the work of moving the existing FastBoot specific funtionality in various addons into `fastBoot/app`. As part of building FastBoot assets faster, we will deprecate the `ember fastBoot` command.
+
+We will also be upgrading [FastBoot](https://ember-fastboot.com) website of how this is works with `ember-cli`.
+
+# Drawbacks
+
+Following are the drawbacks of this approach:
+- [ ] It builds FastBoot specific code in `ember-cli`. This is only being done since we do not want to come up with a generic solution until we fully know the usecases to make this generic that any additional asset can be built.
+- [ ] This approach of loading additive assets after the main asset heavily depends on the loader.js behavior of "last one" wins always. If this behavior changes in future, this approach will break.
```
nathanhammond, line 71: Major confirm, I see this as blocking landing.

active/0000-build-fastboot-in-ember-cli.md
```diff
@@ -0,0 +1,79 @@
...
+```
+my-addon/
+|-- addon/
+│   |-- utils/
+│   │   |-- foo.js
+│-- app/
+│   │-- instance-initializers/
+│   │   |-- error-handler.js
+│-- fastboot/
+│   │-- app/
```
nathanhammond, line 37: This approach implies strongly that `FastBoot` is only additive or clobbering. You propose `postprocessTree` as a way to get out of that, but I'm thinking that we should come up with a better preprocessor/prostprocessor story that allowed for callbacks at the `treeFor` level instead of at the "giant collection" level.

This makes it _far_ easier to compose with nested dependencies.

nathanhammond (commented):
Getting this up as early as I can for visibility. As a result not perfectly wordsmithed, not entirely constructive. But I intend these to be things for which we can work toward solutions.

active/0000-build-fastboot-in-ember-cli.md
nathanhammond: (Seen later in the RFC.)

kratiahuja: @nathanhammond There is no usecase currently that requires a addon inside another addon to do FastBoot work specifically. In general, the goal of this RFC is to _only_ address the current FastBoot usecases and a very limited usecases.

I am open to discussions about how to get this working for addon inside addon but the way I see it should just work how app namespace is built for addon inside addon.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: @nathanhammond I would like to leave that behavior upto the platform to decide. If we merge the two files, it requires us to fork the app.js such that it works in browser differently and in Node environments differently . In Node environment it will be a merge of the two files while in browser it will just be the original app.js file. In general we want to move away from forking and generating two different assets for the primary purpose that it doesn't really guarantee consistency in both environments correctly.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: @nathanhammond The reason why we need these hooks is we want to patch other things (example dom helper patches based on versions) before we start processing the tree. Currently, `ember-cli` doesn't really distinguish the segments of built files and file types clearly for this hook. In the long run, I would be happy to address this.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: @nathanhammond We don't have a usecase currently that requires this to work for nested addon levels. I would assume the order would be same as what `ember-cli` behavior is for app namespace.

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: How is this blocking landing? I am not sure I understand that. Could you please elaborate a bit more?

active/0000-build-fastboot-in-ember-cli.md
kratiahuja: @nathanhammond See this [issue](https://github.com/ember-fastboot/ember-cli-fastboot/issues/270) tracking that already. My current WIP branch of ember-cli-fastboot fixes this somewhat to allow the engine route map to be read and inlined. This is not something specific that needs to be done in `ember-cli` but in `ember-fastboot` project.

kellyselden:
I read through the spike in https://github.com/ember-cli/ember-cli/pull/6395, and it seems like it is so close to being independent of fastboot, all it needs is a supplied "platform".

`preprocessTree` could change from receiving "app-fastboot" to "app-[platform]"
`postprocessTree` could change from receiving "app-fastboot" to "app-[platform]"
`treeForFastBootApp` could change to `treeForAppPlatform` with a param "type" of "[platform]"

An app or addon could then supply an array of platforms, via ember-cli-build.js or package.json, and ember-cli would look for folders with those names. Then the rest of this RFC would take effect.

active/0000-build-fastboot-in-ember-cli.md
nathanhammond: I mean that I don't think we should rely on this behavior, and that we should design something (similar to our whiteboarding session) which doesn't rely on this. Both @rwjblue and I had the same feedback without consulting each other.

active/0000-build-fastboot-in-ember-cli.md
nathanhammond: I'm a firm believer that in our future story we're going to be doing more and more arbitrary splitting of assets. This has already begun with our Engines story and is only going to get more complicated. In other words, I don't know that I agree with this as a long term goal.

[We need to instead do a better job allowing for this behavior to be tested](https://github.com/trentmwillis/ember-asset-loader/pull/34) but since it will hopefully primarily be a framework level concern we can probably feel relatively confident that it "just works" most of the time.

active/0000-build-fastboot-in-ember-cli.md
nathanhammond: Okay, I consider this to be a shortcoming in Ember CLI that we should address.

active/0000-build-fastboot-in-ember-cli.md
nathanhammond: In my head I had a different mental model when posting this comment, which we discussed today at the meeting. This is one of the open problems with my proposal which we need to solve for.

rwjblue:

Thank you!

PR #88, opened 2016-12-07
Use globally installed package managers

Turbo87:
[Rendered](https://github.com/Turbo87/ember-cli-rfcs/blob/use-global-package-managers/active/0000-use-global-package-managers.md)

### DIFF
--- a/active/0000-use-global-package-managers.md
+++ b/active/0000-use-global-package-managers.md
@@ -1,0 +1,139 @@
+- Start Date: 2016-12-07
+- RFC PR: (leave this empty)
+- Ember CLI Issue: (leave this empty)
+
+# Summary
+
+Rely on `npm` and `bower` being installed globally instead of making them
+costly dependencies of `ember-cli`.
+
+
+# Motivation
+
+We are currently specifying [`npm`](https://github.com/ember-cli/ember-cli/blob/v2.10.0/package.json#L99)
+and [`bower`](https://github.com/ember-cli/ember-cli/blob/v2.10.0/package.json#L40)
+as dependencies of `ember-cli`, which means that every project using
+`ember-cli` will currently have a copy of those tools installed in
+their `node_modules` folder.
+
+The motivation for originally bundling them as dependencies was being able
+to bring a consistent experience to all `ember-cli` users and working around
+known bugs in certain versions of `npm`.
+
+Analysis has shown that those two dependencies are the cause for at least 50%
+of all subdependencies, wasted install time and the disk space in the
+`node_modules` folder. This makes it obvious that it is something we should
+try to avoid if possible.
+
+While initially `npm` and `bower` have been somewhat unstable, things have
+improved lately and at least the common `install` subcommand should be
+considered known territory.
+
+It should be noted that all functionality we currently use of both
+dependencies are:
+
+- `npm install`
+- `npm uninstall`
+- `bower install`
+
+Those features are probably the most commonly used subcommands and as such
+the hopefully best tested ones.
+
+Another advantage of this approach would be that supporting
+[`yarn`](https://yarnpkg.com/) will become much easier as well since we
+could transparently fallback to `npm` in case `yarn` is not available.
+We also wouldn't have to package `yarn` as a dependency either, next to
+`npm` and `bower`.
+
+Yet another advantage is that dropping or deprecating support for `bower`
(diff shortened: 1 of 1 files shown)


webark:
another potential alternative could be adding them to the package.json blueprint rather then the ember-cli package itself. That enables a kind of "opt in/out" fit the user, as well as informing what version ideally would be used. Ember init might also be able to detect which version you have, and optionally include it as well. This might be harder to teach though.

stefanpenner:
Yes not apposed to re-evaluating this as the landscape has changed. We could also lazily install the version we need, if the version provided appears invalid or incompatible.

nathanhammond:
I'm in favor of this. I feel like it will make life much better. There will be teething pains but I think we can get it all sorted out.

nathanhammond:
> auto-detecting `yarn` vs. `npm`

I'm opposed to doing this without explicit opt in. I have both, I use both in different scenarios. Simply having them installed doesn't communicate which was used for setting up the current project or any available links. I'd propose an addition to `.ember-cli` as well as a prompt on `ember init` if `yarn` is present.

nathanhammond:
Capturing from Slack:

> We should likely detect if `bower` is missing then install it during `addBowerPackageToProject`.

Because `ember install some-addon` may otherwise break.

stefanpenner:
@Turbo87 there are some constraints we should meet, but I am fine with a path to de-bundle:

we need to (at-least):

* ensure they are using a known to work version (similarly to how we warn \w node)
* Also invoking via the JS api, prevented many sub-process annoyances with node. It may be possible to mitigate these, but will require care.

kellyselden:
What would we do if we detect system npm is buggy 2.0 version.
 * install npm 3
 * warn
 * not work

If we go with warn, we should be very "verbose" to try and sway a person to install a newer npm BEFORE they file a bug report on ember-cli, otherwise we're back to where we were before.

stefanpenner:
@kellyselden we could do what is appropriate for a given command:

* `ember s` likely doesn't care (no change)
* `ember install` error if the wrong version is installed
* `ember addon` error if the wrong version is installed
* `ember new` error if the wrong version is installed
* etc etc

TL;DR it seems we can *nearly* lazily detect, just before we invoke `npm` et al, if that they are right before.

Although, it is likely annoying to not fail as fast as possible...

nathanhammond:
One caveat: if we do this I feel like we need a way to dynamically update the blacklist of non-working `npm`, `yarn`, and `bower` versions remotely.

- Create an `ember-package-manager-blacklist` node module? That will get out of date if people sit on the old version installed in `node_modules` or don't `yarn upgrade`.
- Some JSON file on some server that we download? Always up-to-date, but "phones home."

nathanhammond:
Proposed modification:

> On first `npm` or `bower` invocation if the present system version is not valid (e.g. `npm@2` or less) or non-existent (e.g. no `bower`) we trigger local installation of the desired version. This also sets us up nicely for `yarn` adoption.

And then dropping discussion of `yarn` from this RFC as we don't intend to address that here. (Maybe save off those sections for a future "adopt Yarn" RFC.)

This addresses all of my open concerns.

rwjblue:
If we do this:

> On first npm or bower invocation if the present system version is not valid (e.g. npm@2 or less) or non-existent (e.g. no bower) we trigger local installation of the desired version.

There is no public API surface change here at all, and this is just an internal refactor. I'd be in favor of this refactoring.

---

That being said, it seems silly to make any sort of public facing change in this space without addressing the yarn migration.  If we decide to move forward with the internal refactorings (essentially closing this RFC as not public API) we will still need to RFC to handle `yarn` support...

nathanhammond:
> essentially closing this RFC as not public API

This is true, we can make the refactoring changes here for just-in-time installation without an RFC. We're _very_ close to being able to do that.

rwjblue:
FYI - `bower` has already been swapped to just in time in https://github.com/ember-cli/ember-cli/pull/6631.

rwjblue:
I think this RFC can be closed (since this is a good internal refactoring, not changing public API, and doesn't need a bunch of internal API design).

@ember-cli/core - Thoughts?

Turbo87:
tl;dr from this weeks team call is as @rwjblue suggested in his last comment. we consider this an internal refactoring and will move forward in using the global NPM if it is v3 or above, on v2 we provide a helpful message telling you how to upgrade your local NPM.

PR #89, opened 2016-12-07
Make addons honour app's transpilation by default

cibernox:
[rendered](https://github.com/cibernox/rfcs-1/blob/control-transpilation/active/0000-control-transpilation.md)

### DIFF
--- a/active/0000-control-transpilation.md
+++ b/active/0000-control-transpilation.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-Change how addon code is transpiled to it honors user preferences by default, but still
+Change how addon code is transpiled so it honors user preferences by default, but still
 allow addons to be in control of the transpilation if they need to.
 
 # Motivation
--- a/active/0000-control-transpilation.md
+++ b/active/0000-control-transpilation.md
@@ -16,7 +16,7 @@ Today that is done through `ember-cli-babel`, an addon that is in the default bl
 that the huge majority of apps use.
 
 This transpilation process has some sensible defaults that work for most people, but also allow
-some use configuration for, by example, disable some transformations enabled by default or enable
+some user configuration to, by example, disable some transformations enabled by default or enable
 some experimental ones.
 
 What is less known is that this configuration only affects the application code (the one in your app
--- a/active/0000-control-transpilation.md
+++ b/active/0000-control-transpilation.md
@@ -1,0 +1,75 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Ember CLI Issue: (leave this empty)
+
+# Summary
+
+Allow consumer apps to control the transpilation not only of application code, but also
+of addon code.
+
+# Motivation
+
+At the moment addons control their own transpilation process. This is nice for having a
+"just works" experience, but leaves users without control over the transpiled code they
+ship.
+
+By example, users might want to opt-out from transpiling certain features because
+the application is going to be targeted to fully capable ES6 browsers is still shipping
+ES5 code of their addons, regardless of the babel settings of the app. This leads to
+bundles more verbose than really required.
+
+The generator functions used by ember-concurrency is a good example of an utterly verbose
+transpiled code that users might not want.
+
+The goal of this RFC is to make EmberCLI honor the user's preferences for transpilation (through
+`ember-cli-babel` or perhaps `typescript`).
+
+# Detailed design
+
+The proposed change in behaviour is to apply the user's preferences to the transpilation
+of the `addon` tree instead of the addon's configuration. The addon's configuration is
+honored by their dummy app only.
+For a overwhelming majority of the addons that just use the default settings they won't
+notice anything.
+
+There might be a small number of addons that do customize the settings used by ember-cli-babel,
+and there is where some action is required.
+
+The suggested approach allow addons to access the babeljs configuration so, if they require
+an experimental feature enabled to work, they can throw a warning (or even a hard failure)
+if that feature is not enabled in the host's configuration.
+
+It would be a responsability of the addon to emit those warnings if needed, so ember-cli must provide
+an idiomatic way of accessing the application's configuration. At the moment it is possible
+but is somewhat hacky (`_findHost().options.babel`) and is likely going to change once
+Babel6 is used, which will probably use a `.babelrc` file for configuration instead of the
+current approach as options in the `ember-cli-build.js` file.
+
+# How We Teach This
+
+For end users everything will remain the same.
+For most addon authors it will not require any change. Only those addons that do unusual
+things with babel, like [ember-computed-decorators](https://github.com/rwjblue/ember-computed-decorators/blob/master/ember-cli-build.js)
+may want to do something about it.
+If they don't do anything, things may or may not work depending on the user's settings,
+so it would be sensible for them to inspect the configuration and warn if neccesary.
+
+Checking if a feature is available is probably not trivial, so `ember-cli-babel` should
+expose some helpers that can be required from node-land to verify that the hosts app supports
+a given feature.
+This can also abstract babel5/6 complexities, presents and such.
+
+# Drawbacks
+
+We give the users more control over the transpilation process at the expenses of some complexity
+for addon authors that want to use experimental ES2017 features in their addons, leaving
+up to them to give a good user experience.
+
+# Alternatives
+
+What other designs have been considered? What is the impact of not doing this?
+
+# Unresolved questions
+
+Optional, but suggested for first drafts. What parts of the design are still
+TBD?
--- a/active/0000-control-transpilation.md
+++ b/active/0000-control-transpilation.md
@@ -62,15 +62,14 @@ can attach their own plugins.
 
 #### How can addons be in charge of their own transpilation and yet honor user preferences?
 
-Addons should receive a deep clone of the babel (or typscript) configuration defined in the host app,
-not the original config, so they can safely modify it without affecting the application.
-Each addon receives its own copy so they don't affect other addons either.
+Ember-cli should provide to each addon a deep clone of the configuration defined in the host app
+(by example available in `this.options.babel`), and that config is the one used to transpile the addon.
 
-`ember-cli-babel` will use that object as configuration so user's prerenced are honored unless
-the addon, very deliverately, chooses not to.
+The addon will therefore honor the app's configuration by default, but it can also decide to add
+custom plugins, enable specific features or just throw a warning message if the app's config
+doesn't have required specific transformation enabled.
 
-Making changes in the configuration be scoped to each addon might also avoid some cost where
-some plugin added by one addon is used to transpile code outside that addon.
+By modifying this clone, their changes will not affect the host app or other addons.
 
 # How We Teach This
 
@@ -88,9 +87,13 @@ up to them to give a good user experience.
 
 # Alternatives
 
-What other designs have been considered? What is the impact of not doing this?
+None that I can think of, beside of not doing this.
 
 # Unresolved questions
 
-Optional, but suggested for first drafts. What parts of the design are still
-TBD?
+For this approach it doesn't matter if the version of babel is 5 or 6, however the plugins and their
+names are funamentally incompatible.
+
+- Should addons that depend on `ember-cli-babel` be compiled with their version or with the app's version?
+- If the version of babel in the addon wins but the host apps uses babel6, the config cannot be shared. In
+  that case addons should get the default config?
--- a/active/0000-control-transpilation.md
+++ b/active/0000-control-transpilation.md
@@ -4,46 +4,66 @@
 
 # Summary
 
-Allow consumer apps to control the transpilation not only of application code, but also
+
+Allow consumer apps to control the transpilation config not only of application code, but also
 of addon code.
 
 # Motivation
 
-At the moment addons control their own transpilation process. This is nice for having a
-"just works" experience, but leaves users without control over the transpiled code they
-ship.
+Ember-cli as allowed users to write ES6 code and transpile it to ES5/3 since almost the beginning
+of the project. It's one of those things that people take for granted.
 
-By example, users might want to opt-out from transpiling certain features because
-the application is going to be targeted to fully capable ES6 browsers is still shipping
-ES5 code of their addons, regardless of the babel settings of the app. This leads to
-bundles more verbose than really required.
+Today that is done through `ember-cli-babel`, an addon that is in the default blueprint and
+that the huge majority of apps use.
 
-The generator functions used by ember-concurrency is a good example of an utterly verbose
-transpiled code that users might not want.
+This transpilation process has some sensitive defaults that work for most people, but also allow
+some use configuration for, by example, disable some transformations enabled by default or enable
+some experimental ones.
 
-The goal of this RFC is to make EmberCLI honor the user's preferences for transpilation (through
-`ember-cli-babel` or perhaps `typescript`).
+What is less know is that this configuration only affects the application code. The transpilation
+of code living in addons is done according to the configuration of those addons. While there is
+use-cases for this, particularly addons that want to run their code though a custom babel plugin,
+generally speaking the desired transpiled output is something that the end-users should control.
+
+By example, applications that only target evergreen browsers may want to disable transpilation of
+all ES6 feature since moden browsers already ship them.
+
+This RFC wants to change this behaviour so addons by default honor the host app's preferences,
+while allowing those addons to run their own transformations.
 
 # Detailed design
 
-The proposed change in behaviour is to apply the user's preferences to the transpilation
-of the `addon` tree instead of the addon's configuration. The addon's configuration is
-honored by their dummy app only.
-For a overwhelming majority of the addons that just use the default settings they won't
-notice anything.
+The idea is to still allow addons to be in control of their own transpilation process, but
+make them honour uses preferenced by default.
 
-There might be a small number of addons that do customize the settings used by ember-cli-babel,
-and there is where some action is required.
+There is three different scenarios for addons:
 
-The suggested approach allow addons to access the babeljs configuration so, if they require
-an experimental feature enabled to work, they can throw a warning (or even a hard failure)
-if that feature is not enabled in the host's configuration.
+#### The addon doesn't care about the transpilation process.
+This is the most common common situation. Most addons are authored in ES6 and they don't
+mess with the host app's configuration.
+Those addons should just transpile their code user the host's apps settings.
 
-It would be a responsability of the addon to emit those warnings if needed, so ember-cli must provide
-an idiomatic way of accessing the application's configuration. At the moment it is possible
-but is somewhat hacky (`_findHost().options.babel`) and is likely going to change once
-Babel6 is used, which will probably use a `.babelrc` file for configuration instead of the
-current approach as options in the `ember-cli-build.js` file.
+#### Addons that require experimental features
+There is two options here. Either addons can unconditionally enable that feature only for their
+tree, or it's the user responsability to enable that feature in order to use the addon, with
+optionally the addon doing some feature detection to warn (not force) the user to enable that feature.
+
+By example, an addon that uses `async-await` which right now are only available in Chrome Canary
+could either use the host app's configuration plus this feature or it could just assume the user
+has enabled it and tell them to do so in the docs. Optionally the addon could do feature detection
+(easened by some utilities in `ember-cli-babel` that abstract presets and all that configuration)
+to show a warning in the console.
+
+I believe that the user should have the last word on this, and given the small amount of addons
+that actually do this seems that it's reasonable to update them to behave that way.
+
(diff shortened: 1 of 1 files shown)


active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+
+Today that is done through `ember-cli-babel`, an addon that is in the default blueprint and
+that the huge majority of apps use.
+
+This transpilation process has some sensitive defaults that work for most people, but also allow
+some use configuration for, by example, disable some transformations enabled by default or enable
+some experimental ones.
+
+What is less know is that this configuration only affects the application code. The transpilation
+of code living in addons is done according to the configuration of those addons. While there is
```
Turbo87: to be specific: the app config affects the transpilation of everything inside `app/`, even those `app/` folders inside of addons!

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+
+# Motivation
+
+Ember-cli as allowed users to write ES6 code and transpile it to ES5/3 since almost the beginning
+of the project. It's one of those things that people take for granted.
+
+Today that is done through `ember-cli-babel`, an addon that is in the default blueprint and
+that the huge majority of apps use.
+
+This transpilation process has some sensitive defaults that work for most people, but also allow
```
rwjblue: s/sensitive/sensible/ (I think)

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+
+# Summary
+
+
+Allow consumer apps to control the transpilation config not only of application code, but also
+of addon code.
+
+# Motivation
+
+Ember-cli as allowed users to write ES6 code and transpile it to ES5/3 since almost the beginning
```
rwjblue: `ember-cli` **has**

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+of the project. It's one of those things that people take for granted.
+
+Today that is done through `ember-cli-babel`, an addon that is in the default blueprint and
+that the huge majority of apps use.
+
+This transpilation process has some sensitive defaults that work for most people, but also allow
+some use configuration for, by example, disable some transformations enabled by default or enable
+some experimental ones.
+
+What is less know is that this configuration only affects the application code. The transpilation
```
rwjblue: s/know/known/

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+By example, applications that only target evergreen browsers may want to disable transpilation of
+all ES6 feature since moden browsers already ship them.
+
+This RFC wants to change this behaviour so addons by default honor the host app's preferences,
+while allowing those addons to run their own transformations.
+
+# Detailed design
+
+The idea is to still allow addons to be in control of their own transpilation process, but
+make them honour uses preferenced by default.
```
Turbo87: typo: `user preferences`

active/0000-control-transpilation.md
Turbo87: either `ember-cli` or `Ember CLI`

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+
+By example, applications that only target evergreen browsers may want to disable transpilation of
+all ES6 feature since moden browsers already ship them.
+
+This RFC wants to change this behaviour so addons by default honor the host app's preferences,
+while allowing those addons to run their own transformations.
+
+# Detailed design
+
+The idea is to still allow addons to be in control of their own transpilation process, but
```
rwjblue: Slight tweak suggested here:

> The idea is to still have addons control their own transpilation process, but

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+make them honour uses preferenced by default.
+
+There is three different scenarios for addons:
```
rwjblue: s/is/are/

active/0000-control-transpilation.md
Turbo87: also `as -> has`

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+Those addons should just transpile their code user the host's apps settings.
+
+#### Addons that require experimental features
+There is two options here. Either addons can unconditionally enable that feature only for their
+tree, or it's the user responsability to enable that feature in order to use the addon, with
+optionally the addon doing some feature detection to warn (not force) the user to enable that feature.
+
+By example, an addon that uses `async-await` which right now are only available in Chrome Canary
+could either use the host app's configuration plus this feature or it could just assume the user
+has enabled it and tell them to do so in the docs.
```
Turbo87: or it could use a blueprint that adjusts the Babel settings for the user

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+custom plugins, enable specific features or just throw a warning message if the app's config
+doesn't have required specific transformation enabled.
+
+By modifying this clone, their changes will not affect the host app or other addons.
+
+# How We Teach This
+
+This has no impact for apps, just for addons.
+
+Most addons will not require any change. Only those that want to modify this the transpilation
```
rwjblue, line 69: I do not actually see how ember-computed-decorators needs to be changed **at all** to support this.  If you read through [the recommended snippet](https://github.com/rwjblue/ember-computed-decorators#setup-with-addon) here:

```js
  init: function(app) {
    this._super.init && this._super.init.apply(this, arguments);

    this.options = this.options || {};
    this.options.babel = this.options.babel || {};
    this.options.babel.optional = this.options.babel.optional || [];

    if (this.options.babel.optional.indexOf('es7.decorators') === -1) {
      this.options.babel.optional.push('es7.decorators');
    }
  }
```

You will notice that it already honors any `this.options.babel` settings that are present.

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+
+We give the users more control over the transpilation process at the expenses of some complexity
+for addon authors that want to use experimental ES2017 features in their addons, leaving
+up to them to give a good user experience.
+
+# Alternatives
+
+None that I can think of, beside of not doing this.
+
+# Unresolved questions
```
Turbo87, line 82: we would also have to figure out how Babel resolves plugins. normally they are referenced by name and loaded relative to the Babel package itself, which can be a problem when e.g. `ember-cli-babel` is `npm link`ed.

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+
+The number of addons that does this is even smaller, but one of them is nothing less than Ember Data,
+which uses a babel-plugin to strip Heimdall's instrumentation from production builds.
+
+Since addons remain in control of their transpilation and just reuse the app's configuration, addons
+can attach their own plugins.
+
+#### How can addons be in charge of their own transpilation and yet honor user preferences?
+
+Ember-cli should provide to each addon a deep clone of the configuration defined in the host app
```
rwjblue: We must address how to do this properly with the upcoming babel@5 -> babel@6 transition.  It is possible that this RFC would be blocked on that transition being completed, or only take effect when both app and individual addons are using babel@6.

This needs more thought/design, but I think it is doable...

rwjblue (commented):
I really like this, well done!

active/0000-control-transpilation.md
cibernox: Apparently plugins can also be just `require`d and passed to Babel.

active/0000-control-transpilation.md
cibernox: I think that the idea of deep-cloning the host's app and use that is universal, whatever version of babel we use.

There is an obvious problem if the app uses babel6 and the addon expects babel5, because the options are not compatible between them, but seems a problem of the babel upgrade more than a problem of this RFC per se.

I think that that want to customize the transpilation will have to support both babels at least for a while.

rwjblue: > There is an obvious problem if the app uses babel6 and the addon expects babel5, because the options are not compatible between them, but seems a problem of the babel upgrade more than a problem of this RFC per se.

@cibernox - Right, I totally agree with your assessment here.  However, I am willing to completely block this RFC so that we can focus on the babel@6 upgrade if this will make that migration more complicated.  If we want to ship this sooner, we need to at least address how it doesn't make matters worse (via some sort of mitigation as I suggested above).

active/0000-control-transpilation.md
Turbo87: that is somewhat correct, yes. this does not work inside of a `.babelrc` file though which is JSON-based.

cibernox: I believe that the app should parse its `.babelrc` file and convert it to a JS object. This object is the one that gets deepcloned and passed to addons so they can manipulate it.

I think that for addons that need to mangle the transpilation the `.babelrc` manifest is not expresive enough.

rwjblue: Please note: this RFC does not suggest migration to `.babelrc`

Turbo87: @rwjblue I know, but I'd like us to be aware of that option. We shouldn't build anything that blocks our way of using those files if we want to.

Turbo87:
one thing to keep in mind when thinking about this RFC is https://github.com/babel/babel-preset-env, which I would like to eventually adopt as the default for Babel transpilation in Ember CLI.

active/0000-control-transpilation.md
rwjblue: @Turbo87 - Ya, I completely agree. I wasn't very clear either, I think the RFC **should** address (or at least mention) `.babelrc` and other concepts that we are required to chart through on our road to babel@6 (see other comment in https://github.com/ember-cli/rfcs/pull/89#discussion_r91381533).

active/0000-control-transpilation.md
nathanhammond: Ember CLI

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+Allow consumer apps to control the transpilation config not only of application code, but also
+of addon code.
+
+# Motivation
+
+Ember-cli as allowed users to write ES6 code and transpile it to ES5/3 since almost the beginning
+of the project. It's one of those things that people take for granted.
+
+Today that is done through `ember-cli-babel`, an addon that is in the default blueprint and
+that the huge majority of apps use.
```
nathanhammond, line 16: apps and addons

active/0000-control-transpilation.md
nathanhammond: s/sensitive/sensible

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+
+This transpilation process has some sensitive defaults that work for most people, but also allow
+some use configuration for, by example, disable some transformations enabled by default or enable
```
nathanhammond:
```diff
- , by example, disable
+ disabling

- enable
+ enabling
```

active/0000-control-transpilation.md
nathanhammond: s/know/known/

active/0000-control-transpilation.md
nathanhammond: s/there is/there are

nathanhammond (commented):
Finishing this review since it got updated out from under me.

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+
+The idea is to still have addons control their own transpilation process, but
+make them honour user preferences by default.
+
+There are three different scenarios for addons:
+
+#### The addon doesn't care about the transpilation process.
+This is the most common common situation. Most addons are authored in ES6 and they don't
+mess with the host app's configuration.
+Those addons should just transpile their code user the host's apps settings.
```
nathanhammond: s/user/using/, s/host/root/

active/0000-control-transpilation.md
nathanhammond: I prefer this approach, I think. The addon shouldn't require anything here, it should possibly specify which features it needs support for, and then get venn diagram processing.

Maybe it's cooperative? The root application specifies what its target is, the child addons specify how to achieve that goal?

active/0000-control-transpilation.md
nathanhammond: I would prefer to get to Babel6 first and then land this RFC regardless of any design that demonstrates painlessness. I simply think it's a cleaner migration path less likely to leave skeletons behind.

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+#### How can addons be in charge of their own transpilation and yet honor user preferences?
+
+Ember-cli should provide to each addon a deep clone of the configuration defined in the host app
+(by example available in `this.options.babel`), and that config is the one used to transpile the addon.
+
+The addon will therefore honor the app's configuration by default, but it can also decide to add
+custom plugins, enable specific features or just throw a warning message if the app's config
+doesn't have required specific transformation enabled.
+
+By modifying this clone, their changes will not affect the host app or other addons.
```
nathanhammond: It's theoretically possible that the config defines functions that have scopes and side effects such that the clone isn't valid. We could use vue's global proxy pattern: http://www.2ality.com/2016/11/trace-globals-proxy.html

active/0000-control-transpilation.md
```diff
@@ -0,0 +1,99 @@
...
+# Alternatives
+
+None that I can think of, beside of not doing this.
+
+# Unresolved questions
+
+For this approach it doesn't matter if the version of babel is 5 or 6, however the plugins and their
+names are funamentally incompatible.
+
+- Should addons that depend on `ember-cli-babel` be compiled with their version or with the app's version?
```
nathanhammond, line 87: I'd prefer we only allow this behavior with homogenous versions. If you're not all Babel6 you get current behavior.

nathanhammond:
I think I want this to be cooperative. The addon knows which features it is using, the root application knows what its target is (ES3, ES6, whatever). If the `transpilation target` is understood then the addon should know what it needs to do to get there, starting by building on top of the root application's configuration settings. Which plugins it uses to get there do _not_ communicate exactly what the target is, it specifies only a blacklist (and is thus incomplete).

Moving to this model fits very well with our Broccoli build model where addons are able to build themselves based upon a known target. If we're trying to make the addon guess what to do it will likely always opt for ES3 for max compatibility (I would).

nathanhammond:
This concern is partially alleviated by adopting something like https://github.com/babel/babel-preset-env which @Turbo87 linked us to. If the root app specifies this then the addon knows "what" instead of "how." "How" is simply not enough information to do the right thing 100% of the time.

We should probably encourage adoption of this in root level apps.

/cc @hzoo

cibernox:
@nathanhammond The idea of the app declaring intentions and the addon reacting to those intentions based on the knowledge of the features it uses seems hard.

The app will express is intentions in targets (p.e. **lastest 2 versions of current active browsers**), but addons might express their needs in features (p.e. `async/await`).

I'm not sure how to make them understand each other in a way that it's not madness.

active/0000-control-transpilation.md
```diff
@@ -20,56 +19,48 @@ This transpilation process has some sensible defaults that work for most people,
...
+[smoke-and-mirrors](https://github.com/runspired/smoke-and-mirrors/blob/master/index.js#L27-L42).
+
+However, an undesired side effect of this approach is that, as it stands today, it means that
+the configuration of the user is ignored by addons. That means addons will still transpile features
+(p.e. generator functions) even if the users have disabled them in their apps because they targets
+only evergreen browsers.
 
-By example, applications that only target evergreen browsers may want to disable transpilation of
-all ES6 feature since moden browsers already ship them.
+With ES6 being fully implemented in all modern browsers and many ES2016/ES2017 already starting to
```
stefanpenner, line 37: Yup, for this reason. This is quite important to land.

stefanpenner (commented):
This is a good direction for us to take. 

Question:

* Do we need to concern ourselves with "what happens" when an APP and and Addon have incompatible compilation preferences?

cibernox:
> Do we need to concern ourselves with "what happens" when an APP and and Addon have incompatible compilation preferences?

I'd say we don't. The only incompatibility I can think about is an addon enforcing the transpilation of something the user doesn't want to transpile. I think the addon config always win, although I would consider that addon a *bad boy* for not obeying its parent. But it's not worse than what we have already.

If we are talking about the app providing babel6 config and the addon expecting babel5 config, that is a harder question, but I think that the answer belongs more to the Babel5 -> 6 upgrade rather than this RFC. 

Can some addons be transpiled by babel5 and other by babel6? /cc @Turbo87

active/0000-control-transpilation.md
stefanpenner: Im not sure.

Allowing the system to progressively transition has its benefits. For example, as add-ons upgrade more and more of the users code-gen would no longer need to cater to the lowest common denominator (ES3).

The downside to this, is you may have some ES3 safe code, mixed with your ES6 safe code. Which seems ok.

This would allow our users to see incremental progress. Especially if the most popular / largest add-ons (like ember-data or even ember) made the change more quickly.

The alternative, requiring homogenous setup. Requires the entire universe to transition before progress is seen.

I suspect coupling incremental + a Quest issue, would enable the best outcomes. (e.g. incremental rewards for our community)

Turbo87:
> Can some addons be transpiled by babel5 and other by babel6? 

That was the original plan and it would be very beneficial in the transition period. Otherwise you end up in a situation where some addons only work with older Babel and some only with Babel 6 and you'd have to choose between them...

stefanpenner:
@cibernox and myself had a quick call:

Summary (from @cibernox's #ember-cli-dev summary of the call):

```
[8:41 AM]  
the outcome of the hangout is that giving too much power to users for very little benefit can make transitions harder so the idea would be:
1. Make `babel-preset-env` be part of the babel6 upgrade
2. Just provide addons with the `preset` configuration of the parent app, which typically consist on only a list of target browsers, expressed as `chrome latest` or `> 2%`

[8:42]  
I agree on that, I need to check if this still allows users to plug their own transformations, like ember-data does (edited)

[8:43]  
if so, the RFC could pivot towards "Make `babel-preset-env` a default and make addons obey that by default"
```

active/0000-control-transpilation.md
hzoo: Just as reference there is a PR to add a `.js` to babelrc https://github.com/babel/babel/pull/4892

create-react-app and others do a `require.resolve`: https://github.com/facebookincubator/create-react-app/blob/ab3913cc113b2392fa0329f699571bae3c213308/packages/babel-preset-react-app/index.js#L15

hzoo:
[babel-preset-env](https://github.com/babel/babel-preset-env) only covers preset-latest (no experimental plugins) so it would be the best default behavior but yeah output depends on browser support

cibernox:
@hzoo now that you're here, one of the open point to discussion of this RFC is how addons get access to the babeljs configuration. Addons must be able to customize how they transpile, by example to add their own plugins that perform things like remove intrumentation in production builds and things like that or enable experimental ones if they use stuff that is not yet in the specs, enable loose mode for perf reasons and such.

At the same time, we want to give to the addon the least amount of configuration possible to do their job. @stefanpenner suggested that they should get only the configuration for `env`.

By example, if the configuration in a project is 
```json
{
  "presets": [
    ["env", {
      "targets": { "chrome": 52 },
      "modules": false,
      "loose": true
    }]
  ]
}
```

The addon would only have access to:

```json
{
  "targets": { "chrome": 52 },
  "modules": false,
  "loose": true
}
```

Users can add their own custom transformations having access to just this bit of the configuration?

hzoo:
preset-env itself doesn't include custom transforms so you would have to allow them to append to the preset/plugins array at the top level of the config then? preset-env would only be able to whitelist/blacklist plugins [already included](https://github.com/babel/babel-preset-env/blob/master/package.json#L20) if we add those options.

cibernox:
So, the configuration of a project with custom plugins looks like this:

```js
{
  "presets": [
    ["env", {
      "targets": { "chrome": 52 },
      "modules": false,
      "loose": true
    }]
  ],
  "plugins": ["plugin1", "plugin2", require('./lib/my-plugin')]
}
```

@stefanpenner that makes me think that we might need to pass the configuration pristine to the addons, or alternatively provide a hook/object that gives the user access to **only**  the `env` entry and the `plugins` array.

cibernox:
@rwjblue After ember-cli-babel exists beta, this RFC is essentially implemented, right?

rwjblue:
yes, but not the specific API's suggested IIRC

rwjblue:
@cibernox - I _think_ we can close this one, can you confirm (or update to reflect the current state of things)?

cibernox:
@rwjblue Can you confirm that addons that use babel6 inside babel6 applications with targets also use those targets? That is the only bit I'm not sure if it's implemented.

cibernox:
As of `ember-cli-babel@6.0.0`, this RFC has been implemented. In a few days will be released in the stable channel.

PR #91, opened 2016-12-15
Promote buildInstrumentation to public API

hjdivad:
[rendered](https://github.com/ember-cli/rfcs/blob/addon-instrumentation-experimental-hooks/active/0000-addon-instrumentation-experimental-hooks.md)

- [x] collapse hooks to `instrumentation(type, { summary, tree })`
- [x] update types to `[init, command, build, teardown]`
- [x] update details re command, teardown
- [x] rebuild tree teardown  -> before build instead of after build

### DIFF
--- a/active/0000-addon-instrumentation-experimental-hooks.md
+++ b/active/0000-addon-instrumentation-experimental-hooks.md
@@ -1,0 +1,412 @@
+- Start Date: Wednesday, 14 December 2016
+- RFC PR: 
+- Ember CLI Issue: 
+
+# Summary
+
+Add an instrumentation hook that is available to addons.  This enables users to
+write addons that do things like summarize and report build performance
+information.
+
+- see https://github.com/ember-cli/ember-cli/issues/6349 for additional context.
+- see https://github.com/ember-cli/ember-cli/pull/6606 for an experimental
+  implementation.
+
+# Motivation
+
+Build performance is important to users.  We want to enable users to:
+
+1. Easily discover which portions of their build are costly;
+2. Be able to summarize and report build information in an addon;
+3. Be able to write addons that analyze build performance instrumentation so
+   that they can more easily help diagnose build performance issues in projects
+   to which they do not have direct access.  This is of particular interest to
+   @ember-cli/core &c.
+
+In order to provide these hooks to enable iteration and experimentation prior to
+making firm commitments to format, this rfc propose to initially expose them as
+experiments (see the experiments section below).
+
+# Detailed design
+
+## Experiments
+
+Experiments live in `lib/experiments/index.js`.  Unlike feature flags, there is
+no need to strip them from production.  Experiments allow us to provide power
+user features that are not fully stable without their resorting to private API
+usage.
+
+Experiments are available only in canary builds.  This is achieved by only
+including `lib/experiements/index.js` in canary, and making it the entry point
+for all experiments.
+
+## Instrumentation Hook
+
+We have already a build instrumentation hook as an
+experiment in https://github.com/ember-cli/ember-cli/pull/6546
+
+A more encompassing instrumentation hook is implemented in
+https://github.com/ember-cli/ember-cli/pull/6606
+
+
+The goal of this RFC is:
+
+  1. To make the concept of experiments supported and explicit
+  2. To promote this particular experiment to public API
+
+### Enabling Instrumentation
+
+Instrumentation is enabled if either the environment variable `BROCCOLI_VIZ` is
+set to `1` or if `EMBER_CLI_INSTRUMENTATION` is set to `1`.
+
+If `BROCCOLI_VIZ=1` then in addition to instrumentation hooks being invoked, a
+serialized form of the instrumentation information is written to disk, that is
+appropriate for consumption by [broccoli-viz](https://github.com/ember-cli/broccoli-viz) which is the current behaviour.
+
+### Instrumentation
+
+#### Hook
+
+An addon that implements `instrumentation` will have this hook invoked when
+instrumentation is enabled.
+
+```js
+module.exports = {
+  name: 'my-great-addon',
+
+  instrumentation(name, payload) {
+    // format of instrumentation payload outlined below
+  }
+};
+```
+
+##### name
+
+The `name` argument indicates what phase the instrumentation payload describes.
+In beta and released versions this will always be a string.
+
+On canary it could be a symbol from `lib/experiments` if we add more phases (eg
+more fine-grained phases) for instrumentation information.
+
+The initial set of phases this RFC advocates are:
+  - `init`
+  - `command`
+  - `build`
+  - `shutdown`
+
+##### payload
+
+`payload` is an object with two properties, `summary` and `graph`.
+
+##### payload.summary
+
+The exact format of `payload.summary` depends on the specific phase for which
+the instrumentation hook was called.  In each case, the keys listed are the
+minimum keys that are guaranteed to be present, but there is no guarantee that
+additional information might not also be present.
+
+###### init
+
+`init` covers the period up to, but not including, command execution.  This
+means it's mostly dealing with `require` time.
+
+For `init`, the summary object has the following shape.
+
+```js
+{
+  totalTime,
+  platform: {
+    name,
+  },
+}
+```
+
+- `summary.totalTime` The total time spent during `init`
+- `summary.platform.name` The value of `process.platform`
+
+###### build
+
+`build` covers the time spent in an individual build or rebuild.
+
+For `build`, the summary object has the following shape.
+
+```js
+{
+  build: {
+    type,
+    count,
+    outputChangedFiles
+
+    // additional fields for rebuilds
+    primaryFile,
+    primaryFileCount,
+    changedFiles
+  },
+  platform: {
+    name,
+  },
+  output,
+  totalTime,
+  buildSteps,
+}
+```
+
+- `summary.build.type` one of `'initial'` or `'rebuild'`
+- `summary.build.count` the number of the build (0 for initial build, > 0 for
+  rebuilds).
+- `summary.build.outputChangedFiles` an array of paths to output files that
+  changed during this build.  These paths are relative to the `dist` directory.
+- `summary.build.primaryFile` only present for rebuilds.  Indicates the first
+  file the watcher noticed had changed.
+- `summary.build.changedFileCount` only present for rebuilds.  The number of
+  files the watcher had noticed changed before the build started.
+- `summary.build.changedFiles` only present for rebuilds. The first 10 files
+  the watcher had noticed changed before the build started.
+- `summary.platform.name` The value of `process.platform`
+- `summary.output` The temp directory containing the results of the build.
+- `summary.totalTime` The total time (in nanoseconds) of the build.
+- `summary.buildSteps` The number of broccoli nodes built in this tree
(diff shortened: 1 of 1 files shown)


hjdivad:
Need to revise a couple things after feedback from @rwjblue

hjdivad:
updated

stefanpenner:
👍  (but I am biased, as being one of the implementors :P)

hjdivad:
@stefanpenner nice, now if we can get @rwjblue to 👍 we'll have complete bias coverage.

active/0000-addon-instrumentation-experimental-hooks.md
```diff
@@ -0,0 +1,373 @@
...
+#### Hook
+
+An addon that implements `buildInstrumentation` will have this hook invoked when
+instrumentation is enabled.
+
+```js
+module.exports = {
+  name: 'my-great-addon',
+
+  buildInstrumentation(buildInstrumentation) {
```
kratiahuja: What happens if addons using canary builds of `ember-cli` are used in apps that are using a stable release of `ember-cli`. Do these hooks result in a warning or something like this?

active/0000-addon-instrumentation-experimental-hooks.md
```diff
@@ -0,0 +1,373 @@
...
+- `summary.platform.name` The value of `process.platform`
+
+##### loadTree
+
+The `loadTree` API is exactly the same as the `buildTree` API.
+
+
+## Additional Comments
+
+# How We Teach This
```
kratiahuja, line 388: I know this is something not used by all addon authors regularly but it would be nice to have an example of how to use the API in a real world scenario in the API docs and on ember-cli website. It helps users of `ember-cli` understand the use of this public API better.

active/0000-addon-instrumentation-experimental-hooks.md
rwjblue: In general, we cannot make putting a `buildInstrumentation` method on an addon in older versions of ember-cli throw, but I have implemented a sample that shows how to know that the hook is running ([see here](https://github.com/rwjblue/ember-cli-build-visualizer/blob/309ed77ba78c249085643e26a7b3cf44f6568399/index.js)) by emitting the total number of build steps to the console. This is a totally silly example, just showing that it is possible to use the experiment and properly know it is or isn't functioning (the example prints a `"Nice message here!"`).

active/0000-addon-instrumentation-experimental-hooks.md
rwjblue: Yes, agreed.  We should include adding a section to `https://ember-cli.com/extending/` and and include API docs inline (for `https://ember-cli.com/api/`).

active/0000-addon-instrumentation-experimental-hooks.md
hjdivad: @kratiahuja one of the two goals of this PR is to make the `buildInstrumentation` hook publicly available so it *would* run in stable releases.

Today, to use these features we need a hook that is not `buildInstrumentation`, but rather `require('ember-cli/lib/experiments').BUILD_INSTRUMENTATION` which won't get run at all by current beta or relased versions of ember-cli.

I don't think it makes sense to issue warnings for experiments: part of the motivation of only enabling them in canary and gating them within `lib/experiments` is to make it clear to addon authors what's going on.  It seems odd to issue warnings to end users for this.

What do you think?

active/0000-addon-instrumentation-experimental-hooks.md
hjdivad: @kratiahuja i think that's a great suggestion.  I think the ember-cli website already links to the perf guide, but getting up an example of how to do this in a way that's more discoverable seems like a great idea to me 👍

active/0000-addon-instrumentation-experimental-hooks.md
kratiahuja: I agree it doesn't make sense to issue warnings for releases which don't have it. But I want to make sure that we don't accidentally silence errors on canary builds which are meant to expose these APIs. For example, let's say `require('ember-cli/lib/experiments')` fails on canary build and as @rwjblue example wraps that in a try-catch, we might end up silencing the error. 

Maybe I am overthinking this problem ?

rwjblue: In the case of that example, it doesn't matter if it fails (for any reason) because we will be warning if the hook doesn't get called.  

A better example addressing this specific concern is (used in ember-engines):

```js
var ADDON_TREE_CACHING = (function() {
  try {
    return require('ember-cli/lib/experiments').ADDON_TREE_CACHING;
  } catch (e) {
    if (e.code === 'MODULE_NOT_FOUND') {
      return null;
    }

    throw e;
  }
}());
```

In this case, the only error swallowed is that the module is missing (which is what we would expect for older ember-cli versions).

hjdivad:
Updated per feedback.

hjdivad:
@rwjblue @stefanpenner in particular interested in feedback re: API changes tree -> DAG

active/0000-addon-instrumentation-experimental-hooks.md
```diff
@@ -0,0 +1,400 @@
...
+
+##### payload.graph
+
+`graph` is an object that represents the instrumentation information we have
+gathered for the build.  It is a DAG, whose flow is inverted from the broccoli
+graph. It has a single source node (currently `TreeMerger (all trees)`).
+`payload.graph` is this single source node.
+
+Each node in the graph provides an API for iterating its subgraph as well as
+iterating its own stats The specific nodes in the graph will change over time as
```
rwjblue: Missing a `.` between `its own stats` and `The specific nodes`.

active/0000-addon-instrumentation-experimental-hooks.md
```diff
@@ -0,0 +1,400 @@
...
+One alternative is the status quo: with `BROCCOLI_VIZ=1` users can output a file
+with a similar format that they can post-process offline.  Although this works
+for manual analysis, it is considerably more cumbersome for any automated system
+(such as ongoing monitoring of build performance).  It also does not include
+instrumentation outside of the build, most notably startup.
+
+# Unresolved questions
+
+- [heimdalljs-tree](https://github.com/heimdalljs/heimdalljs-tree) supports
+  `Symbol.Iterator`; should we commit to this as part of our API?
```
rwjblue, line 411: Seems ok to mention (and consider public API), but honestly using `.dfsIterator()` seems much more explanatory (and easier to search)...

rwjblue (commented):
Open questions:

Should we allow enabling a subset of the instrumentation, without enabling others?  For example, should it be possible to turn on `init` instrumentation without enabling the others?  I don't feel very strongly either way, so I suppose we should probably just leave it as "all on" or "all off" until we have a need for that functionality?

--- 

The usage of `adjacentIterator` seems odd to me. Per the [glossary of graph theory terms](https://en.wikipedia.org/wiki/Glossary_of_graph_theory_terms) "adjacent" means:

> The relation between two vertices that are both endpoints of the same edge.

In this provided example:

```
//  TreeMerger
//    |- Babel_1
//    |--|- Funnel
//    |- Babel_2
```

When operating on the node representing `Babel_1`, I would expect `adjacentIterator()` to yield `Babel_2`. It seems like the current documentation is describing something that may more appropriately be named `childIterator`?  The definition of `child` (from the same [glossary of terms](https://en.wikipedia.org/wiki/Glossary_of_graph_theory_terms)):

> In a rooted tree, a child of a vertex v is a neighbor of v along an outgoing edge, one that is directed away from the root.

Given the same example as above, I would expect `.childIterator()` when called on `Babel_1` to yield `Funnel`.

I think we should have both `.childIterator()` (or `childrenIterator`?) and `.adjacentIterator()` (implemented in https://github.com/heimdalljs/heimdalljs-graph/pull/10, pending review/confirmation),  thoughts?

hjdivad:
> Should we allow enabling a subset of the instrumentation, without enabling others? For example, should it be possible to turn on init instrumentation without enabling the others? I don't feel very strongly either way, so I suppose we should probably just leave it as "all on" or "all off" until we have a need for that functionality?

I agree.  Let's leave it all or nothing until someone actually has a need for more fine-grained control.

hjdivad:
> The usage of adjacentIterator seems odd to me. Per the glossary of graph theory terms "adjacent" means:

I think we came to a consensus about this out of band right?

active/0000-addon-instrumentation-experimental-hooks.md
hjdivad: Yeah this is why I'm a little torn.  On the one hand, when you want to visit the entire subgraph, it seems nice to do

```js
for (n of graph) {
}
```

but I agree that in most cases being explicit about the iteration strategy is good

active/0000-addon-instrumentation-experimental-hooks.md
hjdivad: English.  Not even once.

rwjblue:
Yes, we definitely did. I just needed some help thinking things through. 

Basically, `adjacent` nodes are direct parent/children (I had misinterpreted the use of `edge` in the snippet above).

active/0000-addon-instrumentation-experimental-hooks.md
```diff
@@ -0,0 +1,400 @@
...
+
+##### name
+
+The `name` argument indicates what phase the instrumentation payload describes.
+In beta and released versions this will always be a string.
+
+On canary it could be a symbol from `lib/experiments` if we add more phases (eg
+more fine-grained phases) for instrumentation information.
+
+The initial set of phases this RFC advocates are:
```
kratiahuja, line 91: Can we also add some explanation when each of these phases are going to be invoked during the build lifecycle? I am more confused around when is `init` v/s `build` invoked. Are any of the phases overlappping?

kratiahuja (commented):
Looks good to me overall!

active/0000-addon-instrumentation-experimental-hooks.md
hjdivad: This is a great suggestion.  Quick summary for now

1. init is everything before actually running the command.  This means mostly `require`ing time, addon discovery &c.
2. command is the entire command, whatever the command is.  An important exception for `ember serve` is that we'll only include perf information about the last build, to avoid leaks.
3. build is the builder portion of a build command, or of a build within a serve command.
4. shutdown is everything from command to process exit.

hjdivad: Updated, thanks @kratiahuja!

kratiahuja: @hjdivad : Thanks for the summary! Now the full RFC is crystal clear to me 👍 

#3 would also be the last build for serve command right?

hjdivad: @kratiahuja instrumentation will be called for each `build`.  So 3. will be invoked repeatedly.

nathanhammond:
After discussion in today's Ember CLI team meeting we're moving this RFC to FCP. Over the following week please provide your last detailed review to make sure that you're on board with the direction of this new API inside of Ember CLI. Barring significant concerns that result in non-superficial changes to this RFC we intend to land it following the next Ember CLI team meeting one week from today.

Thanks everybody for your work thus far! We're excited to have this new API.

PR #92, opened 2016-12-18
blueprint remove old files

kellyselden:
[Rendered](https://github.com/kellyselden/ember-cli-rfcs/blob/blueprint-remove-old-files/active/0000-blueprint-remove-old-files.md)

Reference PR https://github.com/ember-cli/ember-cli/pull/6472

### DIFF
--- a/active/0000-blueprint-remove-old-files.md
+++ b/active/0000-blueprint-remove-old-files.md
@@ -1,0 +1,48 @@
+- Start Date: 2016-12-17
+- RFC PR: (leave this empty)
+- Ember CLI Issue: (leave this empty)
+
+# Summary
+
+Give blueprint generators the ability to clean up old files.
+
+# Motivation
+
+We want to eliminate the noise of having old files laying after updating
+ember-cli using `ember init`.
+
+# Detailed design
+
+We'd like a declarative API for blueprints to delete files instead of only
+create. It would be essentially syntactic sugar for removing the file yourself
+in an `afterInstall` hook. It would be an array on the blueprint's `index.js`.
+
+```js
+// ember-cli/bluprints/blah/index.js
+module.exports = {
+  // ...
+
+  oldFilesToRemove: [
+    'brocfile.js',
+    'LICENSE.MD',
+    'testem.json'
+  ]
+};
+```
+
+# How We Teach This
+
+The guides could use this addition in the blueprints section, but I envision it
+being used by mostly power users.
+
+A changelog entry should be sufficient to teach this.
+
+# Drawbacks
+
+The only reason to not do this is to hold out for a large blueprint reworking.
+We would be locked into this API.
+
+# Alternatives
+
+The key name can be bikeshed. I chose `oldFilesToRemove` to be verbose and
+explicit, but it can be changed.
--- a/active/0000-blueprint-remove-old-files.md
+++ b/active/0000-blueprint-remove-old-files.md
@@ -13,20 +13,22 @@ ember-cli using `ember init`.
 
 # Detailed design
 
-We'd like a declarative API for blueprints to delete files instead of only
+We'd like an API for blueprints to delete files instead of only
 create. It would be essentially syntactic sugar for removing the file yourself
-in an `afterInstall` hook. It would be an array on the blueprint's `index.js`.
+in an `afterInstall` hook. It would be a returned array on the blueprint's `index.js`.
 
 ```js
 // ember-cli/bluprints/blah/index.js
 module.exports = {
   // ...
 
-  oldFilesToRemove: [
-    'brocfile.js',
-    'LICENSE.MD',
-    'testem.json'
-  ]
+  get oldFilesToRemove() {
+    return [
+      'brocfile.js',
+      'LICENSE.MD',
+      'testem.json'
+    ];
+  }
 };
 ```
 


active/0000-blueprint-remove-old-files.md
```diff
@@ -0,0 +1,48 @@
...
+We'd like a declarative API for blueprints to delete files instead of only
+create. It would be essentially syntactic sugar for removing the file yourself
+in an `afterInstall` hook. It would be an array on the blueprint's `index.js`.
+
+```js
+// ember-cli/bluprints/blah/index.js
+module.exports = {
+  // ...
+
+  oldFilesToRemove: [
```
stefanpenner: It is possible to want to programmatically choose which files are to be deleted? For example, one may want to remove files only if present, or if a specific add-on is present or configuration is set. Given that, should this be declarative or imperative? (I'm not trying to say it one or the other, if possible declarative would be my preference).

active/0000-blueprint-remove-old-files.md
```diff
@@ -0,0 +1,48 @@
...
+# Motivation
+
+We want to eliminate the noise of having old files laying after updating
+ember-cli using `ember init`.
+
+# Detailed design
+
+We'd like a declarative API for blueprints to delete files instead of only
+create. It would be essentially syntactic sugar for removing the file yourself
+in an `afterInstall` hook. It would be an array on the blueprint's `index.js`.
```
stefanpenner: How do the deletion + addition rules compose?

For example:

1. default blueprint: adds `.jshintrc`
2. eslint blueprint:  removes `.jshintrc`


## Scenarios:

### user runs the eslint blueprint for the first time, output would be:

```sh
file remove: .jshintrc
```


### user then later runs `ember init`

One approach would result in:

```sh
file add: .jshintrc
--- eslintrc addon
file remove: .jshintrc
```

One approach would result in no output

```sh
```

One may output exactly what happened:

```sh
file ensured remove: .jshintrc
```

active/0000-blueprint-remove-old-files.md
kellyselden: It makes sense to me to dual function this property as taking an array or a function that returns an array. Is that what you're talking about?

active/0000-blueprint-remove-old-files.md
kellyselden: I was thinking it would be as basic as possible, and let git handle anything more complicated.
* eslint removes .jshintrc
* git commit
* some time passes
* ember init adds it back
* git tells you this file you don't want was added back, so you manually remove the file

If we started composing, I think there would be many corner cases with a lot of complexity. My goal is to get old files out of the project that until now were invisible to you. With my steps above, it is now visible to you that it was removed for some reason, even though there are some extra steps to get rid of it again.

active/0000-blueprint-remove-old-files.md
stefanpenner: its mostly because it is currently a property on the class, and not the instance. Which is a tad funky (although nice for simple addon mains)

kellyselden: I don't follow. Where do you suggest it be moved to?

stefanpenner: As this this implementation will not land before Jan 1, 2017. I believe we can use ES6 feature:

```js
module.exports = {
  get existingFilesToRemove() {
    return [ /* ... */ ]
  }
}
```

This should enable both a reasonable API, and ensure the property exists on the instance and not the class.

stefanpenner: yup confirmed node v4.x supports what we would need (if this approach is taken)

```sh
node-set v4 
  ✓ node.current = v4.6.2
s/s/broccoli-viz ╍{} node
> a = { get foo() { return []; }}
> a.foo # => []
```

kellyselden: I will update.

rwjblue: @stefanpenner - That is also available in 0.12 (its just ES5 getters no?).

screenshot

stefanpenner: @rwjblue yup!

kellyselden:
@stefanpenner I've addressed concerns. This is ready for more scrutiny, or approval. Possible bikeshedding on naming though.

cc @ember-cli/core

active/0000-blueprint-remove-old-files.md
```diff
@@ -0,0 +1,50 @@
...
+# How We Teach This
+
+The guides could use this addition in the blueprints section, but I envision it
+being used by mostly power users.
+
+A changelog entry should be sufficient to teach this.
+
+# Drawbacks
+
+The only reason to not do this is to hold out for a large blueprint reworking.
```
stefanpenner, line 44: I wouldn't worry about the blprnt extraction. That will take some more effort due to how the test harnesses work. And i don't mind backporting any work your doing there rather then blocking the world because i haven't finished that yet.

active/0000-blueprint-remove-old-files.md
```diff
@@ -0,0 +1,50 @@
...
+# Motivation
+
+We want to eliminate the noise of having old files laying after updating
+ember-cli using `ember init`.
+
+# Detailed design
+
+We'd like an API for blueprints to delete files instead of only
+create. It would be essentially syntactic sugar for removing the file yourself
+in an `afterInstall` hook. It would be a returned array on the blueprint's `index.js`.
```
stefanpenner, line 18: So today, file mutations are intended to be transactional. Specifically, it is possible cancel / say no abort a blueprint and abort. As described, this appears like it may not participate in that, is that the case?

kellyselden: I lied then 😛. It is not sugar for `afterInstall` after all, only in concept, because I believe I am still participating in the transactional system here: https://github.com/ember-cli/ember-cli/pull/6472/files#diff-3b6ee55eb49134b6b858dea4972b3117R736

active/0000-blueprint-remove-old-files.md
```diff
@@ -0,0 +1,50 @@
...
+
+# Drawbacks
+
+The only reason to not do this is to hold out for a large blueprint reworking.
+We would be locked into this API.
+
+# Alternatives
+
+The key name can be bikeshed. I chose `oldFilesToRemove` to be verbose and
+explicit, but it can be changed.
```
Turbo87, line 50: @kellyselden `oldFilesToRemove` indicates that this should only be used for files there were previously part of the blueprint. what about other scenarios where files should be removed? e.g. eslint removing jshint files? I'm wondering if that key should be more generic

kellyselden: Great point. One main goal of this RFC is to settle on the best name.

kellyselden:
This was discussed in the Ember CLI team meeting, and there are no more pressing concerns. This is now in Final Comment Period for the next 7 days.

active/0000-blueprint-remove-old-files.md
```diff
@@ -0,0 +1,50 @@
...
+- RFC PR: (leave this empty)
+- Ember CLI Issue: (leave this empty)
+
+# Summary
+
+Give blueprint generators the ability to clean up old files.
+
+# Motivation
+
+We want to eliminate the noise of having old files laying after updating
```
les2, line 11: nit: grammar: "having old files `lying around` after"

les2 (changes requested):
grammar

les2:
How would you register one of those files that include tokens in them

Maybe the oldFilesToRemove should be allowed to have tokens (same tokens as the ones available in the blueprint's files directory).

Edit: ~~Nevermind. I was thinking about something else. :)~~

I was thinking about the correct thing. The question still remains.

Example: A file like `some/thing/__name__.js` could be in the `oldFilesToRemove` if it was moved to `some/other-place/__name__.js`?

rwjblue:
@les2:

> Maybe the oldFilesToRemove should be allowed to have tokens (same tokens as the ones available in the blueprint's files directory).

Agreed, but this RFC doesn't prevent that (and it should be pretty easy to add in a followup PR).

rwjblue:
This has been in final comment period for quite a bit, and no new complications have been raised.

Thank you all for reviewing and contributing!

PR #93, opened 2017-01-06
Multiple Build Targets

kratiahuja:
[Rendered](https://github.com/kratiahuja/rfcs/blob/serve/active/0000-building-target-assets.md)

Due to branching issues, this is a reopen of RFC #75.

### DIFF
--- a/active/0000-building-target-assets.md
+++ b/active/0000-building-target-assets.md
@@ -1,0 +1,187 @@
+- Start Date: 2016-10-31
+- RFC PR: (leave this empty)
+- Ember CLI Issue: (leave this empty)
+
+# Summary
+
+This RFC gives Ember developers to build other platform assets in addition to building the standard browser assets. It is an optional feature and should not affect the build times or the CLI usability for apps that only intend to build browser assets. To limit the scope of this RFC, this RFC only focusses on creating and building targets.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


les2:
@kratiahuja I have a suggestion for your consideration.

*Suggestion:* Place all targets under a common `targets/` folder (or `ember-cli-targets` if `targets/` is likely to cause problems with existing project layouts).

This would offer the following benefits:
1. Group all the targets logically together (except for the default target, which isn't part of the targets hierarchy).
2. Keep target names from conflicting with other folder names
3. Allow auto-discovery of targets by ember-cli.

Example:
```
targets/
    |
    |-- fastboot/
    |
    |-- mobile/
    |
    |-- electron/
    |
    |-- roku/
```

les2:
Update to https://github.com/ember-cli/rfcs/pull/93#issuecomment-274283680

-  Custom `environment.js` config for a target could be located in the `targets/$name/config/environment.js` file. The inherited targets should be merged in and passed as the `defaults` parameter.

The `ember g target` command would create this file by default when generating the target.

rwjblue:
@kratiahuja - Where are we on this? Is this still a thing that we are going to try to push forward?

kratiahuja:
@rwjblue Not for FastBoot but I believe other platforms like electron could benefit from this. However, there is no strong usecase to build this at this time, so my recommendation is to not push forward with this.

rwjblue:

Thank you!

PR #95, opened 2017-01-10
Standarize targets

cibernox:
[rendered](https://github.com/cibernox/rfcs-1/blob/standarize-targets/active/0000-standarise-targets.md)

### DIFF
--- a/active/0000-standarise-targets.md
+++ b/active/0000-standarise-targets.md
@@ -4,8 +4,8 @@
 
 # Summary
 
-This RFC proposes the introduction of a official convention to specify the targets of an
-application.
+This RFC proposes the introduction of a official convention to specify the target browsers
+and node versions of an application.
 
 # Motivation
 
@@ -14,35 +14,38 @@ versions every few weeks. Browsers auto update and each update brings new langua
 and APIs.
 
 Developers need an easy way to express intention that abstracts them from the ever-changing
-landscape of versions and feature matrices, so this RFC proposes to introduce a universal
-configuration option to let developers express their intended targets
+landscape of versions and feature matrices, so this RFC proposes the introduction of a unique
+place and syntax to let developers express their intended targets that all addons can use,
+instead of having each addon define it in a different way.
 
 This configuration should be easily available to addons, but this RFC doesn't impose
-any mandatory behavior on addons. All addons that want to customize their behavior
-depending on the target platforms will have a single source of truth to get that
+any mandatory behavior on those addons. All addons that want to customize their behavior
+depending on the target browsers will have a single source of truth to get that
 information but it's up to them how to use it.
 
 The advantage of having a single source of truth for the targets compared to configure this
 on a per-addon basis like we do today is mostly for better ergonomics and across-the-board
 consistency.
 
-Addons that would benefit from this conventions are `babel-preset-env`, `autoprefixer`,
-`stylelint` and `eslint` (vía `eslint-plugin-compat`), but not only those. Even Ember itself could,
+Examples of addons that would benefit from this conventions are `babel-preset-env`, `autoprefixer`,
+`stylelint` and `eslint` (vía `eslint-plugin-compat`) and more. Even Ember itself could,
 once converted into an addon, take advantage of that to avoid polyfilling or even taking
-advantage of some DOM API (`node.classList`?) deep in Glimmer's internals.
+advantage of some DOM API (`node.classList`?) deep in Glimmer's internals, helping the goal
+of Svelte Builds.
 
 # Detailed design
 
-What seems to be the most popular tool and the state of the art on this is the [browserlist](https://github.com/ai/browserslist)
-npm package.
+What seems to be the most popular tool and the state of the art on building suport matrices 
+for browser targets is the [browserlist](https://github.com/ai/browserslist) npm package.
 
 That package is the one behind `babel-preset-env`, `autoprefixer` and others, and uses the data from
 [Can I Use](http://caniuse.com/) for knowing the JS, CSS and other APIs available on every browser.
 
-The syntax is also pretty flexible as it allows all sort of complex queries like `Firefox >= 20`,
-`>2.5% in CA` (browsers with a market share over 2.5% in Canada) and logical combinations.
+The syntax of this package is natural but also pretty flexible, allowing complex 
+queries like `Firefox >= 20`, `>2.5% in CA` (browsers with a market share over 2.5% in Canada) 
+and logical combinations of the previous.
 
-The way this library works is calculating the minimum common denominator on a per-feature basis.
+The way this library work is by calculating the minimum common denominator support on a per-feature basis.
 
 Per example, if the support matrix for an app is `['IE11', 'Firefox latest']` and we have a linter
 that warns us when we use an unsupported browser API, it would warn us if we try to use
@@ -50,22 +53,29 @@ pointer events (supported in IE11 but not in Firefox), would warn us also when u
 in firefox but not in IE) and would not warn us when using `MutationObserver` because it is supported by both.
 
 This library is very powerful and popular, making relatively easy to integrate with a good amount of
-tools that use it in little time.
+tools that already use it with low effort.
 
-This configuration must be made available to addons but we it's up to the addon authors to take advantage
+This configuration must be made available to addons but it's up to the addon authors to take advantage
 of it.
 
 ### Browser support
 
-The configution of target browsers must in a file that allows javascript execution and exports an object
-with the configuration. Putting the configuration in a file with javascript execution is important because
-allows us to have different config in development and production without having to different keys per 
-environment or any other environment variable.
+The configution of target browsers must be placed in a file that allows javascript execution and exports an object
+with the configuration. The reason to prefer a javascript file over a JSON one is to allow users to 
+dinamically generate different config depending on things like the environment they are building the app in or
+any other environment variable.
 
-One option is the `.ember-cli` file. A new dedicated file under `/config` could also 
-be a good option, much like `config/ember-try.js`.
+One possible location for this configuration is the `.ember-cli` file. 
+A new dedicated named `/config/targets.js` also seems a good option, similar way how addons use `config/ember-try.js`
+to configure the test version matrix.
 
-To addons this list of target browsers will be available in `this.project.targets.browsers`. E.g:
+Ember CLI will require this file when building the app and make the configuration available to addons
+in a `this.project.targets` property. 
+
+This `targets` object contains a getter named `browsers` that returns the provided configuration or the default
+one if the user didn't provide any.
+
+Example usage:
 
 ```js
 module.exports = {
@@ -85,7 +95,7 @@ In addition to browsers, Ember apps can run in node using ember-fastboot, but th
 by the `engines` property in the package.json of the app (or the host app in the case of engines), so no new syntax needed
 for defining the support.
 
-The minimum supported version of node will be available in `this.project.targets.node`. 
+This configuration will be available in the same `this.project.targets` object, but on the `node` property.
 
 ```js
 module.exports = {
--- a/active/0000-standarise-targets.md
+++ b/active/0000-standarise-targets.md
@@ -10,21 +10,21 @@ application.
 ### Motivation
 
 Javascript and the platforms it runs on are moving targets. NodeJS and browsers release new
-versions every few weeks. Browsers auto update and each update bings new language features
+versions every few weeks. Browsers auto update and each update brings new language features
 and APIs.
 
-Developers need an easy way to express intention that abstracts them from the everchanging
+Developers need an easy way to express intention that abstracts them from the ever-changing
 landscape of versions and feature matrices, so this RFC proposes to introduce a universal
 configuration option to let developers express they intended targets
 
 This configuration should be easily available to addons, but this RFC doesn't impose
-any mandatory behaviour on addons. All addons that want to get customize their behaviour
+any mandatory behavior on addons. All addons that want to get customize their behavior
 depending on the target platforms will have a single source of truth to get that
 information but it's up to them to do it.
 
 The advantage of having a single source of truth for the targets compared to having several
 configurations, each one with its own `targets` is mostly for better ergonomics and across-the-board
-consitency.
+consistency.
 
 Addons that would benefit from this conventions are `babel-preset-env`, `autoprefixer`,
 `stylelint` and `eslint` (vía `eslint-plugin-compat`), but not only those. Even Ember itself could,
@@ -77,7 +77,7 @@ module.exports = {
 In addition to browsers, Ember apps can run in node using ember-fastboot. If that is the
 case, we need to cover node as a target, but it deserves a different entry on `targets.node`.
 
-The sintax for for this configuration option is even simpler, since apps will only run in a single
+The syntax for for this configuration option is even simpler, since apps will only run in a single
 version of node and developers are in control of it, so the only thing users must specify is the
 minimum supported version of node.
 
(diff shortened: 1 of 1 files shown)


cibernox:
This RFC is somewhat an extraction of RFC in PR #89, but less controversial in the sense that focuses on standarizing app targets and its syntax but leaves open how ember, ember-cli or addons should use them.

I open this in hope that it can be approved faster and independently of the advances in the babel6 migration.

active/0000-standarise-targets.md
```diff
@@ -0,0 +1,152 @@
...
+
+# Summary
+
+This RFC proposes the introduction of a official convention to specify the targets of an
+application.
+
+### Motivation
+
+Javascript and the platforms it runs on are moving targets. NodeJS and browsers release new
+versions every few weeks. Browsers auto update and each update bings new language features
```
twokul: `bings` -> `brings`?

cibernox: yup

active/0000-standarise-targets.md
```diff
@@ -0,0 +1,152 @@
...
+```
+
+### Node support
+
+In addition to browsers, Ember apps can run in node using ember-fastboot. If that is the
+case, we need to cover node as a target, but it deserves a different entry on `targets.node`.
+
+The syntax for for this configuration option is even simpler, since apps will only run in a single
+version of node and developers are in control of it, so the only thing users must specify is the
+minimum supported version of node.
```
Serabe: If this is added, I would either let the user specify the node version as they do in the `package.json` or specify in the name that is the minimum version acceptable. The first option is harder to implement, but people would be more used to it.

active/0000-standarise-targets.md
```diff
@@ -0,0 +1,152 @@
...
+
+While this RFC standardizes a concept that will open the door to better and more comprehensive tooling,
+it makes us choose one syntax (the one used by [browserlist](https://github.com/ai/browserslist)) over
+any other perhaps superior choice that may exist or appear in the future.
+
+# Alternatives
+
+Let every addon that wants to deal with targets to have a `targets`-like option in its configuration
+instead of standardizing a single configuration option, which effectively leaves things as they are
+right now.
```
Serabe, line 121: This option opens the door of a misconfiguration, in the sense that you might add support for X browser in autoprefixer but not in babel and suddenly, your life is much more interesting. :)

cibernox: I didn't say it was a good alternative 😉 But this is the state of things right now: each addon uses it's own configuration, and possibly its own syntax.

stefanpenner:
big fan of this direction!

I'll try and fine time over the next few days to thoroughly review.

cibernox:
@stefanpenner any opinion on core people on this? I could work in the implementation myself. I'm craving for some tooling.

active/0000-standarise-targets.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Motivation
+
+Javascript and the platforms it runs on are moving targets. NodeJS and browsers release new
+versions every few weeks. Browsers auto update and each update brings new language features
+and APIs.
+
+Developers need an easy way to express intention that abstracts them from the ever-changing
+landscape of versions and feature matrices, so this RFC proposes to introduce a universal
+configuration option to let developers express they intended targets
```
rwjblue: s/they/their/

active/0000-standarise-targets.md
```diff
@@ -0,0 +1,152 @@
...
+depending on the target platforms will have a single source of truth to get that
+information but it's up to them how to use it.
+
+The advantage of having a single source of truth for the targets compared to configure this
+on a per-addon basis like we do today is mostly for better ergonomics and across-the-board
+consistency.
+
+Addons that would benefit from this conventions are `babel-preset-env`, `autoprefixer`,
+`stylelint` and `eslint` (vía `eslint-plugin-compat`), but not only those. Even Ember itself could,
+once converted into an addon, take advantage of that to avoid polyfilling or even taking
```
rwjblue: it is an addon as of ember-cli@2.11 / ember-source@2.11.0 🎉

active/0000-standarise-targets.md
```diff
@@ -0,0 +1,152 @@
...
+advantage of some DOM API (`node.classList`?) deep in Glimmer's internals.
+
+### Detailed implementation
```
rwjblue: Not sure if this was intentionally left blank?

active/0000-standarise-targets.md
```diff
@@ -0,0 +1,152 @@
...
+```js
+var app = new EmberApp(defaults, {
+  targets: {
+    browsers: ['>2%', 'last 3 iOS versions', 'not ie <= 8']
+  }
+});
+```
+
+This configuration must be made available to addons. It's up to the addons to use it.
+I suggest to make it available in `app.targets`
```
rwjblue: I would prefer one of these instead:

* add `this.targets` property
* `this.project.targets()` / `this.project.targets`

active/0000-standarise-targets.md
```diff
@@ -0,0 +1,152 @@
...
+The syntax is also pretty flexible as it allows all sort of complex queries like `Firefox >= 20`,
+`>2.5% in CA` (browsers with a market share over 2.5% in Canada) and logical combinations.
+
+Using this syntax would allow us to integrate easily with other tools.
+
+Example usage:
+
+```js
+var app = new EmberApp(defaults, {
+  targets: {
```
rwjblue: We should avoid tying this target system to the build time only `ember-cli-build.js` / `EmberApp` instance since it will be needed from places outside of the build environment. For example, when editing a project I enjoy utilizing my editors built in `eslint` system to indicate any linting errors/warnings that I have in my code.  If I wanted to utilize `eslint-plugin-compat`, I would have to duplicate the configuration in `ember-cli-build.js` and `.eslintrc.js` (or change `ember-cli-build.js` to have multiple exports).

I'd prefer that we either:

* put the config in `.ember-cli` file which is already in the project
* put the config in another file in `config/` (e.g. `config/supported-targets.js`)

active/0000-standarise-targets.md
```diff
@@ -0,0 +1,152 @@
...
+I suggest to make it available in `app.targets`
+
+```js
+module.exports = {
+  name: 'ember-data',
+
+  included(app) {
+    this._super.included.apply(this, arguments);
+
+    console.log(app.targets.browsers); // ['>2%', 'last 3 iOS versions', 'not ie <= 8']
```
rwjblue: The meaning of `app` changes when using things like fastboot and ember-engines, this is why I would like to decouple storing of target from `app`...

active/0000-standarise-targets.md
rwjblue: I'd suggest that the default would just do `node: require('../package').engine` (deferring to the engine listed in `package.json`).

active/0000-standarise-targets.md
```diff
@@ -0,0 +1,152 @@
...
+If this property is not provided, addons can assume that this app does not target node.
+
+# How We Teach This
+
+This is a new concept in Ember CLI, so guides will have to be updated to explain this
+concept. The good part is that this new concept can help enforcing with tools a task were
+traditionally enforced only with peer reviews.
+
+To ease the transition Ember CLI can also, in the absence of a specific value provided by the user,
+default to a predefined matrix of browsers that matches the browsers officially supported by the framework.
```
rwjblue, line 103: 👍 - But we should (in this RFC) list out the "default" configuration...

rwjblue (commented):
> but this RFC doesn't impose any mandatory behavior on addons

I agree that we shouldn't (or can't) mandate anything, but we should recommend that they honor them...

active/0000-standarise-targets.md
cibernox: I'm also ok with that. We can reuse it.

active/0000-standarise-targets.md
cibernox: It wasn't blank, it's just that it was an h3 and it should be an h1, and the next points are subsections.

active/0000-standarise-targets.md
```diff
@@ -0,0 +1,152 @@
...
+
+The syntax for for this configuration option is even simpler, since apps will only run in a single
+version of node and developers are in control of it, so the only thing users must specify is the
+minimum supported version of node.
+
+Example usage:
+
+```js
+var app = new EmberApp(defaults, {
+  targets: {
```
stefanpenner: targets object, to be consider together? e.g. lowest common denominator between all properties, or is this 2 separate builds... I believe they should be `lowest common denominator`

rwjblue:
Discussed during ember-cli core team meeting today, we were generally very 👍.  Main issues are:

* naming ("target" is hard to use, perhaps we can do somethign like `target-platforms`, `supported-platforms`, etc...)
* determine exactly where to put this config (IMO it should not be in `ember-cli-build.js`, but that is TBD obviously)
* confirm if multiple targets are supposed to be considered as a combination (e.g. "lowest common denominator")
* list the explicit default config that matches ember-source@2.11's supported platforms

active/0000-standarise-targets.md
rwjblue: aha! thank you!

active/0000-standarise-targets.md
cibernox: Actually, one interesting approach of this is the one `eslint-plugin-compat` uses. It reads the `browserlist` property in the package.json.

If we are cool with reading the `engines` property of the package.json for the node support it might also be a reasonable place to the list of supporter browsers.

active/0000-standarise-targets.md
cibernox: This is affectively how things like `babel-preset-env` and `eslint-plugin-compat` works. Browserlist has an evergreen database from caniuse.com with features by browser version, so it calculates the support on a per-feature basis.

I'll add it.

active/0000-standarise-targets.md
cibernox: I'm not sure what Ember supports. I know that IE support starts in 9, but no idea about the policy for other evergreen browsers. Is there one? Who knows it?

kellyselden:
If the default is dependent on which ember version you may be using, should it be bundled with ember (like its blueprints)? I'm not sure what this would look like in a way you could override it in ember-cli land...

active/0000-standarise-targets.md
cibernox: Added support matrix from the browsers we test in saucelabs & travis. No mention to ios tho, but likely IE9 will be the weakest ring of the chain in everything anyway.

cibernox:
@kellyselden It could be part of ember-cli, because a ember drops browser support in major versions only, which would carry (I think) a major version bump in ember-cli as well.

kellyselden:
@cibernox True, but we still allow ember 1.13 in ember-cli 2.x, and will be the same for the next cycles. Might be unnecessary friction to have more code in cli that ties to specific ember versions.

cibernox:
@kellyselden So options:

- Make it a part of ember-source, like blueprints. In the absence of configuration, we use those values.
- We trust the file (e.g `config/targets.js`). This file is part of the default blueprint of an app, and lives wherever blueprints live now (I think they've extracted). If the file is not present, we don't have any default value.

I lean towards the first. At the end ember.js already has (kind of) this information about browser support in the config files for travis and saucelabs.

active/0000-standarise-targets.md
```diff
@@ -42,9 +42,18 @@ That package is the one behind `babel-preset-env`, `autoprefixer` and others, an
...
-This configuration must be made available to addons. It's up to the addons to use honor it.
+Per example, if the support matrix for an app is `['IE11', 'Firefox latest']` and we have a linter
+that warns us when we use an unsupported browser API, it would warn us if we try to use
+pointer events (supported in IE11 but not in Firefox), would warn us also when using `fetch` (supported
+in firefox but not in IE) and would not warn us when using `MutationObserver` because it is supported by both.
+
+This library is very powerful and popular, making relatively easy to integrate with a good amount of
+tools that use it in little time.
+
+This configuration must be made available to addons but we it's up to the addon authors to take advantage
```
ro0gr: odd "but we"

cibernox:
Feedback applied:

- Per slack bikeshedding, the configuration will be in `this.project.targets`. The object in that property will have `browsers` and `node` properties. There's another RFC  #93 that wants to introduce another kind of targets. The chamption (@kratiahuja) seems happy to use `this.project.targets.platforms` for its concept.

- Added data of where the config must be placed. Preferred option: `config/targets.js`, similar to `config/ember-try.js`, which is also used to build a support matrix, but of ember versions instead of browser versions.

- Confirmed Minimum Common Denominator of the approach, and added an example.

- Added list default supported browsers, extracted from the browsers tested in ember in travis and saucelabs.

kanongil:
Does the `node` property encompass supporting alternative engines, like [node-chakracore](https://github.com/nodejs/node-chakracore)?

cibernox:
@kanongil The node property is basically `require('package.json').engines`. Do you know if node-chakracore is specified that way?

rwjblue:
Discussed in ember-cli core meeting today, we are 👍 on advancing this to final comment period.

One thing that we would like to see before landing the RFC (after FCP is finished and whatnot) is a spike of the implementation in a PR to ember-cli.  This is just to confirm that all of the constraints listed here are addressed and we don't find flaws.  

I believe that I will have time to spike this out in the next couple of days...

cibernox:
Nice! If you don't let me know and I can try.

Cryrivers:
speaking of targets, is there any consideration for targets like Cordova, Electron, or "Ember Native" (maybe in the future)?

rwjblue:
@Cryrivers - Yes, but that is a somewhat separate RFC (see https://github.com/ember-cli/rfcs/pull/93 for some background).

cibernox:
@Cryrivers @rwjblue Actually after some investigation there is a small overlapping on the concepts. Some libraries like `babel-preset-env` consider electron a target, just like any other browser, and use https://github.com/kilian/electron-to-chromium to translate from the version of electron to the equivalent chromium version to know if a feature is supported.

The targets this RFC targets are the javascript runtimes in which this app is going to run. That is the browser or electron version, or even node version in fastboot. 

A separate concept is "platforms" (name to be bikeshedded) that only makes sense when you are going to wrap your app and you are going to write javascript code that bridges to the OS using electron/cordoba APIs. 

In summary, in this RFC a target is "where my Ember app code is going to run on". The electron specific code is probably going to live in a separate folder and RFC #93 wants to solve that. It's an area I don't have any expertise on, TBH.

kanongil:
I'm a bit concerned that the `project.targets.browsers` naming is a bit too generic. I would prefer that it signals that it only applies to the `javascript` runtime. Maybe just prefix is with `"js"` so that it becomes `project.jsTargets.browsers `?

As it is, you can confuse it with other runtime capabilities, like css capabilities, which e.g. https://github.com/kimroen/ember-cli-autoprefixer handles.

rwjblue:
@kanongil - I believe that is entirely the point.  We want it to be used for *both* JS transpilation and CSS transpilation targeting. This is mentioned in the **Motivation** and **Detailed Design** sections of this RFC.

cibernox:
@kanongil exactly. It's not scoped to js. CSS is part of the plan. My example if my app drops support for IE11 then my `ember-css-linter` (invented addon) can warn me wherever I use `-ms-user-select: none`, since I don't need the prefix anymore.

cibernox:
I have a working implementation of this https://github.com/ember-cli/ember-cli/pull/6776
It was pretty simple to implement. 

I've also [spiked](https://github.com/machty/ember-maybe-import-regenerator/pull/3) how ember-maybe-import-regenerator (used by ember-concurrency) could use it to conditionally avoid including the regenerator polyfill if the built matrix supports generator functions.

rwjblue:
Discussed in ember-cli core team meeting yesterday, and with some small (scope narrowing) changes we are ready to land this.

Specifically, we would like to restrict `targets`to only include `browsers` at this point (IOW, remove `node` section).  A future RFC can add support for things like `fastboot` / `cordova` / etc, and removing `node` here shouldn't affect the primary motivators of the RFC.

@cibernox - Let me know when you have had a chance to update that, and we are ready to land this...

cibernox:
@rwjblue Amended the RFC to leave node/fastboot out of this RFC, so will cross that bridge another day.

kanongil:
In case you missed it, I posted some comments on javascript issues in #104.

PR #98, opened 2017-02-27
dynamic import syntax transformation in Ember CLI

frederikbosch:
This RFC proposes implementing a dynamic import syntax transformation in Ember CLI.

### DIFF
--- a/active/0000-dynamic-import-syntaxt.md
+++ b/active/0000-dynamic-import-syntaxt.md
@@ -1,0 +1,27 @@
+- Start Date: (fill me in with today's date, 2017-02-27)
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes implementing a dynamic import syntax transformation in Ember CLI.
+
+# Motivation
+
+Recently, the proposal for adding a "function-like" import() module loading syntactic form to JavaScript has
+reached stage 3 of the TC39 process. Other asset pipeline tools like
+Webpack [already support the `import()`](https://github.com/airbnb/babel-plugin-dynamic-import-webpack) syntax.
+
+For Ember users it can be quite useful to dynamically import modules. This is well explained in the proposal.
+
+> This could be because of factors only known at runtime (such as the user's language), for performance reasons
+(not loading code until it is likely to be used), or for robustness reasons (surviving failure to load a non-critical
+module). Such dynamic code-loading has a long history, especially on the web, but also in Node.js (to delay startup
+costs). The existing import syntax does not support such use cases.
+>
+>Truly dynamic code loading also enables advanced scenarios, such as racing multiple modules against each other and choosing the first to successfully load.
+
+# Implementation
+
+Once Ember CLI [supports Babel 6](https://github.com/ember-cli/ember-cli/issues/5015), it should be quite easy to
+implement this RFC By using the [Babel plugin to transpile `import()` to `require`](https://github.com/genkgo/babel-plugin-dynamic-import-amd).
--- a/active/0000-dynamic-import-syntaxt.md
+++ b/active/0000-dynamic-import-syntaxt.md
@@ -120,4 +120,4 @@ export default Ember.Service.extend({
 
 # Future
 
-Now that Ember is still concats javascript files, this RFC does not effect network performance now. However, if Ember decides to stop concattenating files (when HTTP 2 is available in every setup), this RFC could be emtremely useful to import only those modules that are needed.
+Now that Ember still concatenates javascript files, this RFC does not effect network performance now. However, if Ember decides to stop concatenating files (when HTTP 2 is available in every setup), this RFC could be extremely useful to import only those modules that are needed.
--- a/active/0000-dynamic-import-syntaxt.md
+++ b/active/0000-dynamic-import-syntaxt.md
@@ -25,3 +25,47 @@ costs). The existing import syntax does not support such use cases.
 
 Once Ember CLI [supports Babel 6](https://github.com/ember-cli/ember-cli/issues/5015), it should be quite easy to
 implement this RFC By using the [Babel plugin to transpile `import()` to `require`](https://github.com/genkgo/babel-plugin-dynamic-import-amd).
+
+# Examples
+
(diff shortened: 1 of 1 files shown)


locks:
@frederikbosch the structure of this document does not match the current template, did you make sure to update your repository before branching?

frederikbosch:
@locks I took one of the earlier RFCs as template. I guess I missed something...

frederikbosch:
@locks Maybe you can also tell if you like the general idea of this RFC. I love to update, and spend more time on it, if people feel this is heading into the right direction.

stefanpenner:
The general idea is good.

Some Questions:

* what does the on-disk format look like for "lazy loaded modules", different folder or folders?
* how do we teach this
* provide some concrete motivating examples (loading some large `wysiwyg` editor code only in some places)
* how does this related/work with/or contradict https://github.com/emberjs/rfcs/pull/158

frederikbosch:
@stefanpenner I updated the RFC with one concrete example and a list of other possible directions. Since Ember is already using import/export syntax, this RFC will not have a high impact. It only proposes to allow _dynamic_ imports, which is an upcoming javascript feature. Therefore, this should not contradict any other work, rfc or feature. Furthermore, there is no additional work required to teach this. Those who will need to import some module dynamically, they can be redirected to the standard.

frederikbosch:
Extended the number of examples even further and also added a future section

active/0000-dynamic-import-syntaxt.md
```diff
@@ -0,0 +1,123 @@
...
+
+```js
+import Ember from 'ember';
+
+export default Ember.Service.extend({
+  someMethodOnThisService() {
+    return Ember.RSVP.Promise((resolve, reject) => {
+      fetch('/resource').then((data) => {
+        if (data.variable === 1) {
+          import('app/services/x').then(service => resolve(service.default), reject);
```
kratiahuja, line 43: How is this different than doing something like this in an app:

```
   import require from 'require';
   ....
   if (data.variable === 1) {
     require('app/services/x').then
     ...
  }
```

What's the motivation to introduce a new public API when one already exists (Ember's loader API)?

frederikbosch:
@kratiahuja Thanks for your feedback. What's the motivation to use import/export syntax instead of define and require (Ember's loader API) everywhere? That is that Ember is complying to newest JS syntax. Same is true for dynamic imports. It's an upcoming JS feature.

I think eventually Ember's loader API will only exist as polyfill for import/export. Or, when [the official loader API will land](https://whatwg.github.io/loader/) as implementation for that API.

frederikbosch:
With the [recent work by @rwjblue](https://github.com/ember-cli/ember-cli/pull/6828), this RFC might become even more relevant. I would love to create a PR for this RFC if it gets accepted.

rwjblue:
I'm in favor of this, and am happy to push things forward at the next ember-cli core team meeting.

@frederikbosch - Can you please prepare a PR to ember-cli-babel (current master branch) implementing this? Its often quite useful to see an implementation in the wild, to know that an RFC has been properly informed by the implementation details...

Pauan:
Hello. I work for [amCharts](https://www.amcharts.com/). We have recently published a pre-release of the next big version of our software library (version 4).

Our version 4 library uses the dynamic `import` syntax. This works perfectly fine in Webpack, but one of our customers is using Ember, and it does not work for them.

We believe this sort of situation will happen more frequently in the future, as more libraries start using the dynamic `import` syntax.

Pauan:
P.S. We use the dynamic `import` to load npm modules, like this: `import("fabric")`, so if this RFC is accepted Ember should make sure that it also supports dynamic `import` of npm modules.

frederikbosch:
@Pauan The simple (untested) Ember solution to that is the following.

```js
// vendor/dynamic-import.js
window['import'] = function(moduleName) {
  return Ember.RSVP.Promise.resolve(window.require(moduleName));
}

// ember-cli-build.js
app.import('vendor/dynamic-import.js');

Pauan:
@frederikbosch I don't think that will work, because `import("fabric")` is not valid syntax in ES5, so it cannot be polyfilled like that. The transformation *must* happen at build-time.

frederikbosch:
Why not? Dynamic import is a regular function, and the above is a polyfill for it. If `fabric` is added as an Ember plugin or added as vendor in `ember-cli-build.js`, it *will* be transpiled during builds and you should be able to get this to work.

Pauan:
@frederikbosch Dynamic `import` is *not* a regular function. It is special magical syntax, just like top-level `import` and `export`. To be more specific, [`import` is a keyword](https://www.ecma-international.org/ecma-262/6.0/#sec-keywords), and thus it *cannot* be used as a variable or function name.

frederikbosch:
@Pauan You're right, my solution was too simple. I thought that worked, and I still think it did until browsers like Chrome started implementing the feature.

Pauan:
To add a bit more clarification: yes you can use `window["import"] = ...` to *define* an `import` function, however when you *call* `import("foo")` one of two things will happen:

1. In browsers that *don't* support dynamic `import` you will get a SyntaxError, because `import` is a keyword.

2. In browsers that *do* support dynamic `import` you will get a "module could not be found" error.

In either case the `window["import"]` function will *never* be called. Thus, it is impossible to create a polyfill for dynamic `import`.

frederikbosch:
You might also consider using parameters of [ember-cli-babel](https://github.com/babel/ember-cli-babel).

frederikbosch:
And then use the [babel plugin](https://www.npmjs.com/package/babel-plugin-dynamic-import-amd) that I already created.

Pauan:
@frederikbosch Yeah, that might work. I wasn't aware that an AMD version of `babel-plugin-dynamic-import` existed, thanks!

In that case, shouldn't it be very easy for Ember to support dynamic `import`? It can just enable the `babel-plugin-dynamic-import-amd` plugin by default.

frederikbosch:
Yes it is. But it had to go through RFC. And I created this RFC, including with the plugin that should provide the functionality. But then I should also create a PR to `ember-cli-babel`, including tests, and that is I believe what I have not yet.

rwjblue:

Thank you!

PR #99, opened 2017-03-03
Use a wrapping factory to simplify importing external libraries by fixing their missing imports and exports

frederikbosch:
This proposal tries to simplify importing external libraries. More specifically, it tries to fix those libraries that do not use an AMD or CommonJS module loader without having to create an addon.

There is also a personal reason for this contribution. Recently I have been writing my own asset pipeline using `Broccoli`. And I am very happy and grateful with the `Broccoli` ecosystem. I feel I need to share my solutions to problems that might be also valid for Ember CLI. Mainly because those who contribute to `Broccoli` are doing so because of Ember. Moreover, I am also a very happy Ember user. However, it might be true that this proposal is not valid. The time I spend on Ember is not as much as it used to be. So I it might be true that there already is feature for this and therefore this proposal is superfluous. At least I hope this contributes some how.

### DIFF
--- a/active/0000-amd-factory.md
+++ b/active/0000-amd-factory.md
@@ -1,0 +1,98 @@
+- Start Date: (fill me in with today's date, 2016-06-14)
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


rwjblue:

Thank you!

PR #100, opened 2017-03-06
Include ember-cli-dependency-lint in the default app blueprint

dfreeman:
[Rendered](https://github.com/dfreeman/ember-cli-rfcs/blob/4b8997ffbdc23cf32aaeba860ceb8db44ea9aa1a/active/0000-dependency-lint.md)

### DIFF
--- a/active/0000-dependency-lint.md
+++ b/active/0000-dependency-lint.md
@@ -1,0 +1,63 @@
+- Start Date: 2017-03-06
+- RFC PR: (leave this empty)
+- Ember CLI Issue: (leave this empty)
+
+# Summary
+
+Add [ember-cli-dependency-lint](https://github.com/salsify/ember-cli-dependency-lint) to the default app blueprint.
+
+# Motivation
+
+When installing dependencies for an Ember app, it's possible to end up with multiple copies of a given addon if it's
+transitively needed by other addons with conflicting version requirements. When Ember CLI includes that addon into a
+built app, only one version will ultimately make it in, meaning that code written expecting a different version may not
+work properly. Today this situation is hard to detect without manual inspection, and may arise any time you add or
+update a dependency.
+
+# Detailed design
+
+Further details are available in [ember-cli-dependency-lint's README](https://github.com/salsify/ember-cli-dependency-lint),
+but at a high level it validates for each addon in the build that only one version is present, and then exposes any
+violations of that rule in two ways:
+ - as failing lint tests to alert the user to the problem during the normal development flow
+ - via an explicit `ember dependency-lint` command that facilitates faster cycles when troubleshooting a dependency issue
+
+Users may override the "only one version" rule on a case by case basis by specifying a semver expression.
+
+# How We Teach This
+
+Many users today haven't deeply considered the underlying problem that ember-cli-dependency-lint detects. As Ember
+Weekend called out, "it helps you debug an issue you probably didn't know you had".
+
+The problem itself has proven pretty straightforward to grasp once folks are presented with it, but understanding what
+action to take when it rears its head requires a more subtle understanding of how npm packages and dependency resolution
+work than many people seem to have. A quick primer in the Ember CLI documentation may help with this (and actually
+might be a nice add regardless of this RFC), but the exact strategy for this is called out as an Unresolved Question
+below.
+
+_A note on the name:_ I realized right after publishing that ember-cli-dependency-checker is also a thing, and that I'd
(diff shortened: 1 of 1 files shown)


Gaurav0:
Sounds good. Can someone check if there is any performance regression? If not, I'm in favor.

bcardarella:
Isn't this going to be resolved on its own with the new module system?

rwjblue:
No? We aren't changing the way depedencies work with module unification...

dfreeman:
> Can someone check if there is any performance regression?

Is there an existing test harness for verifying perf regressions in the CLI, or are you after something more ad hoc?

> Isn't this going to be resolved on its own with the new module system?

Unless I missed something major in that RFC, I don't think the new module system impacts this. Module unification doesn't change the way addon content makes its way into the build, does it? (I.e. the final build output will still only have a single copy of that addon's code)

Edit: @rwjblue beat me to the punch 🙂

Gaurav0:
@dfreeman I mainly want to know if cold/warm/re build times don't increase significantly with the addon installed. You can benchmark simply by running ember build for cold/warm and ember serve for rebuilds. For more info, see https://github.com/ember-cli/ember-cli/blob/master/PERF_GUIDE.md

dfreeman:
@Gaurav0 Oh I'm well familiar with that guide—just wanted to know how scientific of an answer you were looking for ;)

The addon reads one config file and writes one lint test file, and the rest is just navigating the in-memory project representation that the CLI builds itself, so perf impact should be exactly the same across cold and warm builds. The only cost on rebuild is the fs check that discovers that the tests have already been generated.

I'll double check and get some actual numbers, but my intuition is that the difference between builds with and without the addon is likely to be pretty minimal.

dfreeman:
I did two runs each on a fresh app from CLI 2.11.1, with and without the addon. The perf differences look to be small enough to be in the noise of whatever else was happening on my machine at the time.

<table>
<tr><th></th><th colspan=2>Without the Addon</th><th colspan=2>With the Addon</th></tr>
<tr><td>Cold</td><td>7464ms</td><td>7905ms</td><td>7930ms</td><td>7746ms</td></tr>
<tr><td>Warm</td><td>1200ms</td><td>1164ms</td><td>1165ms</td><td>1156ms</td></tr>
<tr><td>Rebuild</td><td>115ms</td><td>139ms</td><td>104ms</td><td>86ms</td></tr>
</table>

jrjohnson:
When we added this to our app we found two errors.  I found the message to be easily navigable and it was clear exactly where we needed to look in our application and in our dependency tree to find issues. 

The only way this could have been simpler or more useful is if it created the necessary addon PR on its own - but then I would probably be out of a job.  I think adding this is good for application owners and really great for the addon community in general.

rwjblue:
I am in favor of this, and plan to bring it up at the ember-cli core team meeting today...

stefanpenner:
it seems strange to have both this and ember-cli-dependency-checker, should they become the same thing?

dfreeman:
@stefanpenner I have a kind of half memory of @rwjblue (or... someone?) saying in Slack that he saw them as two different things. That said, I don't feel particularly strongly either way, and the names of the two are already similar enough that it's not obvious which does what.

kellyselden:
Agreed. Now that I think of it, those addons sound like they do the same job at face value. I vote for combination, or a name change to differentiate. something like ember-cli-addon-dependency-graph-lint (perhaps exaggerated in descriptiveness).

jrjohnson:
They are different things - re-naming might help differentiate that.  
Combining them would make it impossible to opt out of one without losing the other so I vote for re-naming if one of these needs to happen.

_edited for clarity of opinion_

kellyselden:
> those addons sound like they do the same job at face value

I meant the way their names read without knowing what they do.

dfreeman:
@kellyselden `ember-cli-do-you-have-multiple-versions-of-addons-that-might-cause-problems-except-for-the-ones-where-its-okay-checker`? 😜 

I do like the idea of working the word "addon" in there somewhere though, since it doesn't actually inspect other types of dependencies.

acorncom:
> ember-cli-do-you-have-multiple-versions-of-addons-that-might-cause-problems-except-for-the-ones-where-its-okay-checker

_Folks, I think we have a winner.  But, do you think we might hit issues with max name length on NPM?_

:trollface:

kellyselden:
> Folks, I think we have a winner. But, do you think we might hit issues with max name length on NPM?

Forget NPM, Windows is doomed...

rwjblue:

Thank you!

PR #102, opened 2017-03-13
RFC: Pre-build Addons

trentmwillis:
[Rendered](https://github.com/trentmwillis/rfcs-1/blob/pre-build/active/0000-prebuild-addons.md)

### DIFF
--- a/active/0000-prebuild-addons.md
+++ b/active/0000-prebuild-addons.md
@@ -55,13 +55,19 @@ Pre-built asset bundles will be generated for Ember-CLI's default target as well
 
 During the build of consuming applications, if the target and environment information match one of the prebuilt asset bundles for a given addon, then those assets will be used by the build pipeline in place of the various `treeFor*` hooks that would normally execute for that addon.
 
-If a consumer wishes to build the addons from scratch with their application, they will be able to specify an `ignorePrebuiltAddons` flag in their `ember-cli-build.js`, like so:
+If a consumer wishes to build addons from scratch with their application, they will be able to specify a `prebuiltAddons` options hash in their `ember-cli-build.js`, like so:
 
 ```js
 const app = new EmberApp(defaults, {
-  ignorePrebuiltAddons: true
+  prebuiltAddons: {
+    'cool-addon': false
+  }
 });
 ```
+
+This hash defaults to `true` for all addons and allows you to turn it off for individual addons. If you want to disable all addons, you can specify: `'*': false`.
+
+If your application uses a non-default build target or environment, then the above hash will default `false` and you can opt-in to pre-builds by setting addons to `true`.
 
 # How We Teach This
 
@@ -92,4 +98,3 @@ To limit potential download and file system I/O costs, an alternative format for
 # Unresolved questions
 
 - Per-target installation of addons? Is it possible to publish multiple versions of addons with pre-built assets for different targets/environments? If so, this may fix the concerns over custom pre-builds.
-- For non-default targets, what happens? Are they opted out of all pre-building? Can they opt-in again if they know it is fine for their app?
--- a/active/0000-prebuild-addons.md
+++ b/active/0000-prebuild-addons.md
@@ -1,5 +1,5 @@
 - Start Date: 2017-03-08
-- RFC PR: (leave this empty)
+- RFC PR: [#102](https://github.com/ember-cli/rfcs/pull/102)
 
 # Summary
 
--- a/active/0000-prebuild-addons.md
+++ b/active/0000-prebuild-addons.md
@@ -1,0 +1,95 @@
+- Start Date: 2017-03-08
+- RFC PR: (leave this empty)
+
+# Summary
+
+Allow Ember Addons to be "pre-built" such that they produce assets which can be included more efficiently in the build process of a consuming application.
+
+# Motivation
+
+When working on Ember-CLI applications, it is common to include many addons; this is one of the great benefits of the Ember ecosystem. However, in our current system, addons are built with the application they are included in. This means that for each addon you include in your application, you incur a non-trivial impact to your application's build performance.
+
+Since most addon consumers never customize an addon's build or change its source code, it would be beneficial to provide pre-built versions of their assets to improve build performance. Doing so will benefit all applications and allow larger applications, especially those being developed with Engines (which are packaged as addons), to scale to truly ambitious levels.
+
+# Detailed design
+
+## Pre-builds for Addons
+
+A new command will be introduced for use inside of addons only and used as part of the publishing process:
+
+```bash
+ember pre-build
+```
+
+This command will generate assets into a directory named `prebuilt` in the root of the addon. Within `prebuilt` will be a series of directories representing prebuilt assets for different combinations of targets and environments. Since targets can be non-primitive values, they will be hashed for the corresponding directory path.
+
+```bash
+cool-addon/
+  |-prebuilt/
+    |-48e380a165c417253512a904d6b5cf2b-prod/
+    |-48e380a165c417253512a904d6b5cf2b-dev/
+```
+
+Within each of the prebuilt asset directories, there will be directories which contain the results of each `treeFor*` hook in the addon prior to being merged with a consuming application.
+
+```bash
+cool-addon/prebuilt/48e380a165c417253512a904d6b5cf2b-prod/
+  |-addon/
+  |-addon-templates/
+  |-addon-test-support/
+  |-app/
+  |-public/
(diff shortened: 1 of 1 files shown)


active/0000-prebuild-addons.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Consuming Pre-builds
+
+During the build of consuming applications, if the target and environment information match one of the prebuilt asset bundles for a given addon, then those assets will be used by the build pipeline in place of the various `treeFor*` hooks that would normally execute for that addon.
+
+If a consumer wishes to build the addons from scratch with their application, they will be able to specify an `ignorePrebuiltAddons` flag in their `ember-cli-build.js`, like so:
+
+```js
+const app = new EmberApp(defaults, {
+  ignorePrebuiltAddons: true
```
givanse: how about:

```
const app = new EmberApp(defaults, {
  prebuiltAddons: {
    'ember-cli-foobar': false
  }
});
```

Defaults to true for every addon, but it lets you turn it off for individual addons.

trentmwillis: Good suggestion and this could feed nicely into the opt-back-in story for app's with non-default build targets.

active/0000-prebuild-addons.md
```diff
@@ -0,0 +1,95 @@
...
+
+If additional, common targets/environments become present we can roll them into the preset builds. Alternatively, we can add this feature in the future, but leaving it out for now is much easier than potentially having to unwind it.
+
+## Pre-built Asset Format
+
+To limit potential download and file system I/O costs, an alternative format for the pre-built assets may be considered. Using a JSON representation of the pre-built assets would allow for storage as a single, large file which can reduce the amount of overhead in the pre-builds. This approach, however, makes understandability and debuggability of the pre-built assets much less and should likely be considered after sufficient real-world data has been collected on the current suggested approach.
+
+# Unresolved questions
+
+- Per-target installation of addons? Is it possible to publish multiple versions of addons with pre-built assets for different targets/environments? If so, this may fix the concerns over custom pre-builds.
```
givanse, line 100: npm scoped packages maybe? where the scope is the name of the addon and then
```
// of course ember install would handle this
npm install @ember-cli-foobar/target-name
```

I don't know if Yarn has the same feature, might be a concern.

Publishing becomes more complex, but it seems it can be automated easily.

trentmwillis: This is one potential option. My familiarity here is limited though, so if this is non-trivial to make easy for developers, then it may be worthy of a follow-up RFC to this one.

dfreeman: One downside to using scopes for this is that it doesn't provide a clear path for addons that are already scoped to e.g. an organization.

rwjblue:
I'm generally 👍  here, but would like to see a few things fleshed out a bit more:

* We should add a hook `shouldUsePrebuilt` (or something better) to addons (similar conceptually to `isDevelopingAddon`) which returns `true` or `false` 
* We should add a hook to `prebuiltTreeFor` or something, so that the addon can control the tree returned / used for the prebuild.
* Need to flesh out what the interaction is for in repo addons.  Ideally, they can still take advantage of prebuilding, and "deopt" when a user actually edits the source they represent.  I'm happy to punt the very specific details for this,  but we should think through the general issues that this would face and make sure we have hooks at the proper levels.
* Ideally we at least allow "fetching" the prebuilt asset via alternative methods. This would definitely not be a default (the default is to include the prebuilt output in the `npm pack` output and publish to NPM), but we should have appropriate hooks present so that it is possible to retrieve the prebuilt asset from a different location.

active/0000-prebuild-addons.md
```diff
@@ -0,0 +1,100 @@
...
+## Consuming Pre-builds
+
+During the build of consuming applications, if the target and environment information match one of the prebuilt asset bundles for a given addon, then those assets will be used by the build pipeline in place of the various `treeFor*` hooks that would normally execute for that addon.
+
+If a consumer wishes to build addons from scratch with their application, they will be able to specify a `prebuiltAddons` options hash in their `ember-cli-build.js`, like so:
+
+```js
+const app = new EmberApp(defaults, {
+  prebuiltAddons: {
+    'cool-addon': false
```
stefanpenner, line 63: may also be nice, to turn off all prebuilts via an environment variable.

stefanpenner (changes requested):

active/0000-prebuild-addons.md
```diff
@@ -0,0 +1,100 @@
...
+
+```bash
+cool-addon/prebuilt/48e380a165c417253512a904d6b5cf2b-prod/
+  |-addon/
+  |-addon-templates/
+  |-addon-test-support/
+  |-app/
+  |-public/
+  |-styles/
+  |-templates/
```
ro0gr, line 43: Does templates format matter? What happens if `some-addon` is pre-built with _Glimmer1_ and the consuming application is using _Glimmer{2,3,..}_?

ro0gr: I mean js version of templates generated for `Glimmer1` look something like:
```javascript
Ember.HTMLBars.template((function () {
    return {
      meta: {
        "revision": "Ember@2.8.3",

```
and `Glimmer2`:
```javascript
Ember.HTMLBars.template({ "id": "dxsyziu1", "block": {} // ...
```
Seems like these 2 formats are not backward compatible?

ro0gr:
I think one of the biggest issue here is increasing maintainability/learn-ability aspects for both sides:
 - addon maintainers. Who should take care of pre-builds. 
 - addon consumers. Who should disable/enable pre-built dependencies.

It would be great if we could move the whole responsibility of dependencies pre-building to the `ember-cli` which could cache each non `isDevelopingAddon` dependency just on the first build after `npm i` and persist it across server restarts/dependencies re-installations.

ro0gr:
I've just found something similar in my tmp `/tmp/simple_concat_concat-input_base_path-XXXXXXXX\modules`. It's basically pre-built `addon` directories per each module:
```
- modules/                                        
 |+ ember-ajax/                                   
 |+ ember-cli-page-object/                 
 |+ ember-load-initializers/                 
 |+ ember-resolver/                             
 ...
```
Each module is transpiled to _ES5_+_amd_. 
Looks like `ember-cli` + `broccoli` already have most important pieces of information?

cibernox:
I'm dubious about wether saving a few seconds on a cold start is worth adding all this complexity. Also, this somehow goes a bit against some recent advances in ember-cli.

The `targets` feature available in 2.13-beta allows the user to specify the browsers that the app supports. 

One addon that takes advantage of this is ember-cli-babel, which thanks to babel-preset-env transpiles only what is indispensable for the given targets, and this works all the way down to the addons. The idea is that addons respect user preferences and if all your targets browsers  support arrow functions, the code of your addons should honor your preferences and also avoid transpiling arrow functions.

The same thing applies to CSS (autoprefixer/post-css), conditional inclusion of polyfills, etc...

That means that this addon should know when to NOT use the prebundled version based on...things. To make things more complicated, checking that the targets of the addon and the targets of the app are the same is not good enough, because the way targets are specificed usually includes some option like `last 2 versions of Chrome`. And whatever were the last 2 versions of chrome when the addon was "prebundled" might not be the same versions when the addon is included.

I see a lot of potential for this task to be a huge mess for a (IMHO not very critical) boot time improvement.

mixonic:
I agree with many concerns @cibernox raises.

The RFC discusses applications opting out. Am I wrong than an addon could itself change the build process of an Ember app, and that the cached trees for the addon would be invalid at that point? Can an addon disable prebuilds for other addons?

What if we address a bug in the Ember-CLI build pipeline- how can we ensure cached content that was built with said bug doesn't propagate after we've fixed Ember-CLI itself?

There simply must be more drawbacks if we apply a little creative thinking here. For example what is the impact on the size of the NPM package? Given an app you have access to, what would you expect the change in `node_modules` size to be given that several/all addons would be precompiled?

trentmwillis:
Thanks all for the feedback. I'll be updating the RFC to address concerns and provide more details.

mixonic:
Thanks @trentmwillis 🙇

trentmwillis:
Hey all, after taking the provided feedback into account and discussion with @rwjblue and others, I've decided to withdraw this proposal.

The primary motivation of _providing a mechanism to help build times scale efficiently as a project grows_ remains, but this approach seems to be attempting to provide too high-level of a solution. We're going to explore what functionality and potential hooks are needed to curb build time growth before revisiting.

Again, thanks everyone for the feedback!

PR #107, opened 2017-05-25
enhance `ember install`

kellyselden:
[Rendered](https://github.com/kellyselden/ember-cli-rfcs/blob/ember-install/active/0000-ember-install.md)

### DIFF
--- a/active/0000-ember-install.md
+++ b/active/0000-ember-install.md
@@ -1,0 +1,35 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+
+# Summary
+
+`ember install` without arguments should install all packages.
+`ember install package-name` should not error if the package is not an
+ember addon.
+
+# Motivation
+
+Since Ember CLI can now decide for you whether to use Yarn or not, we would like to
+take advantage of that in more ways.
+
+# Detailed design
+
+This is now possible since we don't have to worry about Bower.
+Instead of you knowing whether you should run `npm install` or `yarn`,
+`ember install` would do that for you.
+Since Ember CLI would now potentially be installing non-ember addons
+when it wasn't before, it makes sense to expand this to
+`ember install package-name` and not error if the package is not an
+ember addon.
+
+# How We Teach This
+
+This can be adding to the guides in the project setup and installing addons sections.
+
+# Drawbacks
+
+I don't think there are any reasons not to do this.
+
+# Alternatives
+
+None


webark:
would there need to be a corresponding "remove/uninstall" then?

kellyselden:
> would there need to be a corresponding "remove/uninstall" then?

No. Since there is already an install command without an uninstall command, that seems out of scope.

buschtoens:
Instead of erroring ember-cli should print a note, indicating to the user that the installed package is not an Ember add-on. 

> The package '${packageName}‘ is not an Ember add-on. In order to use it you'll probably need to perform some further custom setup.

active/0000-ember-install.md
```diff
@@ -0,0 +1,35 @@
...
+`ember install` without arguments should install all packages.
+`ember install package-name` should not error if the package is not an
+ember addon.
+
+# Motivation
+
+Since Ember CLI can now decide for you whether to use Yarn or not, we would like to
+take advantage of that in more ways.
+
+# Detailed design
```
stefanpenner, line 15: This needs to be fleshed out more. As the ember command will attempt to find the local ember, and sad things may happen.

This is worsened by partial node_module checkouts.

Details on how this will be solved would be useful.


One approach would be for `ember install` to always fallback to the global `ember` installation?

stefanpenner (changes requested):

webark:
ya. with the motivation being that ember-cli chooses to use yarn for you.. if you just npm uninstall won't the reference stay in your yarn.lock? i guess it wouldn't actually install the package later.. but just seems a little messy, and that ideally you're still going to have to keep track if your using yarn or npm (same with when pulling a project for the first time)

active/0000-ember-install.md
kellyselden: Wouldn't this just piggy-back on how `ember install package-name` finds the `ember` bin? Why would this RFC be focused on that resolution?

kellyselden:
@webark We already suffer from the problem that `ember install package-name` may modify your `yarn.lock`, but we have no `ember uninstall package-name`. Solving that would be the scope of a different RFC IMO.

active/0000-ember-install.md
stefanpenner: I think that is the trickiest bit to make this successful

Turbo87:
FWIW I'm not in favor of this RFC. `ember install foobar` serves the purpose of installing the dep **and** running the corresponding blueprint, but everything else can just be installed without ember-cli having to care about it. I think making `ember install` and alias for `npm/yarn install` increases the amount of perceived magic done by the Ember ecosystem.

Regarding `ember install` should work for non-addons: I thought that already worked without any issues. If it doesn't I would consider it a bug that should be fixed.

benkingcode:
> No. Since there is already an install command without an uninstall command, that seems out of scope.

I don't understand the reasoning behind this. If the goal is to make the use of `npm` or `yarn` 'invisible' to the end user, why would there not be a command to remove packages? It seems very unintuitive.

rwjblue:

Thank you!

PR #108, opened 2017-06-19
Add API to allow addons to define custom transform

kratiahuja:
[rendered](https://github.com/kratiahuja/rfcs/blob/add-import-api/active/0000-add-custom-transform.md)

cc: @stefanpenner @rwjblue

### DIFF
--- a/active/0000-add-custom-transform.md
+++ b/active/0000-add-custom-transform.md
@@ -1,0 +1,107 @@
+- Start Date: 2017-06-18
+- RFC PR: (leave this empty)
+
+# Summary
+
+This RFC proposes to add a new API to allow addons to register a custom transformation. This transformation can then be used by other addons when calling `app.import` with `using` API.
+
+# Motivation
+
+Addons or apps may want import browser only compatible libraries using `app.import` via bower or npm. These libraries should not be running in Node.
+
+When FastBoot was doing two builds (to generate different assets for browser and Node environment), addon or apps often conditionally imported these libraries relying on the value of `process.env.EMBER_CLI_FASTBOOT`. With the new scheme of the build where only additional Node assets are built, this enviornment is no longer exposed.
+
+In order to expose better semantics to allow apps and addon authors to easily import these libraries without much overhead (see issue [here](https://github.com/ember-fastboot/ember-cli-fastboot/issues/413)), we need to have these libraries wrapped with an FastBoot check. This can be achieved by extending the `using` API of `app.import`. FastBoot addon would like to register a custom transformation that other FastBoot compatible addons may chose to use in a declarative API.
+
+# Detailed design
+
+Today, `ember-cli` supports transforming anonymous AMD modules imported via `app.import` into named AMD modules:
+
+```js
+app.import('/path/to/module.js', {
+  using: [
+    { transformation: 'amd', as: 'some-dep' }
+  ]
+});
+```
+
+The `amd` transform is hardcoded in `ember-cli`. However, it is not possible for addon authors to provide any additional transformation that other addons can use when importing third-party modules. Addons like, FastBoot would like to provide custom transformation for other addons to use so that they can wrap their third party libraries in Node environments.
+
+In order to do this, we would like to expose an API that allows addons to register a custom transformation. This API will be an advanced API and will only be used by addons that want to provide custom transformation. Other addons can chose to use that custom transformation using its name.
+
+The API to register a custom transformation in `ember-cli` will be defined in `index.js` of the addon and will be an advanced API:
+
+```js
+addCustomTransform() {
+  return {
+    'fastboot-shim': function(tree, options) {
+
+      return stew.map(tree, function(content, relativePath) {
+        return `if (typeof FastBoot === 'undefined) { ${content} }`;
+      });
+    }
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


trentmwillis (commented):
Overall, I like this and think it adds a nice bit of extensibility to the `import` API. That said, I think we should use a more specific name than "custom transformation".

Since the transform is only applicable for imports, we should likely refer to them as "import transforms" and the corresponding hook could then be `addImportTransform`. Otherwise, it will likely communicate that these transforms can be run on any code in the build pipeline, which I can see causing confusion.

kratiahuja:
Agreed. I am terribly bad at naming things :)

hjdivad (commented):
I agree with @trentmwillis although I'd also add that the [addon API](https://ember-cli.com/api/classes/Addon.html) uses nouns so it's a bit strange to add a function that starts with `add` but seems to function similarly to existing APIs.

My 2c on the bikeshedding would be for something like `importTransforms`.

---

Minor bikeshedding aside, what behaviour do we want for addons that provide transforms of the same name?  Shall we make this an error case initially?

trentmwillis:
> What behaviour do we want for addons that provide transforms of the same name? Shall we make this an error case initially?

My opinion is that it should either error or, at the least, log a message stating which one is being used. We may also want to consider having an "override" functionality, in case a consumer wants to override a transform that is bundled together with something else (e.g., override the fastboot transform, but still needs the rest of the fastboot addon).

hjdivad:
> My opinion is that it should either error or, at the least, log a message stating which one is being used. We may also want to consider having an "override" functionality, in case a consumer wants to override a transform that is bundled together with something else (e.g., override the fastboot transform, but still needs the rest of the fastboot addon).

Override with a warning is fine as long as we guarantee the order in which addons are processed for custom transforms.

kratiahuja:
For the override functionality, I would actually prefer creating a `from` optional keyword which allows consumers of the transform to define which one they chose to pick. Example an addon `my-addon` provides a slightly different of `fastboot-shim` from `ember-cli-fastboot` and names it `fastboot-shim`:

```js
app.import('path/to/module.js', {
  'using': [{ transformation: 'fastboot-shim', from: 'my-addon'}]
});
```

If an addon importing the third party library choses to not use `from` keyword, then the last definition of `fastboot-shim` wins (based on the order of the addons being run). We can log it in that case. 

I'll update the RFC to account for this case.

hjdivad:
@kratiahuja I'm concerned about adding features before we have an actual need for it.

Is there a reason to prefer the `from` feature over just erroring or overriding + warning for now until we have an actual case of two addons that could reasonably be used together but want to register import transforms with the same name?

Although I suppose it could be argued that guaranteeing addon order with override is a roundabout way of implementing `from`, I would see it more as a "soft error".

The tl;dr is my vote is to not support the case preemptively, but also to not have undefined semantics (hence the preference for hard or soft erroring)

kratiahuja:
> Is there a reason to prefer the from feature over just erroring or overriding + warning for now until we have an actual case of two addons that could reasonably be used together but want to register import transforms with the same name?

Since the ordering is governed by the addons that define the transform, for an app that wants to use the the transform from an addon that has the same name as defined by the other addon becomes extremely cumbersome to do so. They have to ask the addon owner that defined the transform to change the name. I am fine with the default behavior being overridden + warning when `from` not used but I think it is important to provide a complete solution for such edge cases. 

The reason I am hesitant over erroring it is since apps that consume two addons (that provide the same transform name) are now left with the choice of choosing which addon to keep in their list in order to use the correct version/have their build passing. I think it is important to not introduce friction in such cases for apps.

kratiahuja:
Updated per feedback from the June end ember-cli team meeting.

kratiahuja:
@chadhietala Would love your feedback on this RFC wrt if this is going to be an issue with tree shaking. I am told you had some thoughts on `app.import` in general that was discussed in ember-cli core meeting in Portland.

chadhietala:
I believe we have decided that anything that comes through `app.import` is treated as opaque and just concated into a bundle. I don't have issues with adding a transform to wrap the stuff going through `app.import` as long as the input from doing so is held separately from the files that would candidates for shaking.

kratiahuja:
Sure sounds good. Thanks @chadhietala

hjdivad:
We have discussed this during the weekly meetings and decided to move to FCP

kratiahuja:
Implementation of RFC: https://github.com/ember-cli/ember-cli/pull/7245

kratiahuja:
I changed the` importTransforms` API to accomodate `processOptions` functions as well. `processOptions` should be defined if you want to process an additional properties app developers provide along with the transformName. `importTranforms` can either define only the callback or callback and processOptions.

Therefore it can have a structure as when defining `processOptions`:
```js
importTransforms() {
  return {
     'amd': { 
        transform: function(tree, options) {},
        processOptions: function(assetPath, entry, options) {
            // do the processing
            return options;
       }
     }
  }
}
```

When the processOptions does not need to be defined, the API can be used as:
```js
importTransforms() {
  return {
     'amd': function(tree, options) {},
  }
}
```
I rarely see addons requiring to processOptions, therefore wanted to make sure we support both.

rwjblue:
Awesome, thank you for updating on the implementation progress.

Considering that there is no new discussion here, and that implementation has been going well, I think we should merge.

twokul:
🎉

PR #109, opened 2017-07-24
Environment Selection

rwjblue:
[rendered](https://github.com/ember-cli/rfcs/blob/environment-selection/active/0109-environment-selection.md)

Relates to:

* https://github.com/ember-cli/ember-cli/issues/7217
* https://github.com/glimmerjs/glimmer-blueprint/issues/12

And in other communities:

* https://github.com/angular/angular/issues/6189
* https://github.com/facebook/react/issues/8784

### DIFF
--- a/active/0109-environment-selection.md
+++ b/active/0109-environment-selection.md
@@ -1,0 +1,90 @@
+# Environment Selection
+* Start Date: 2017-07-24
+* RFC PR: (leave this empty)
+
+# Summary
+
+The goal of this RFC is to justify and solidify the mechanism that ember-cli uses to determine the environment that any given command will be ran under. 
+
+# Motivation
+
+We have received feedback from our direct users, developer relations folks, and benchmark maintainers suggesting that our current system for selecting the environment does not set our users up for success.
+
+This RFC aims to ensure that users of ember-cli get the correct environment when specified (via `NODE_ENV`, `EMBER_ENV`, or `--environment` command line flags) and when using the default.  
+
+# Terminology
+
+* Command line specified environment -- When the command line invocation includes a specified `environment`  (either `--environment=xyz` or uses the environment shortcut aliases (e.g. `-prod` or `-dev`).
+* ember-cli specific process environment variable specified environment -- When the shell that invokes the command has `process.env.EMBER_ENV` set.
+* General purpose process environment variable specified environment -- When the shell that invokes the command has `process.env.NODE_ENV` set.
+* Test command -- A command used to run tests (e.g. `ember test` or `ember test --server`).
+* Interactive command -- A command which is not intended to "complete". These commands commonly perform rebuilds and/or re-run tests (i.e. `ember build --watch`, `ember serve`). 
+* Non-interactive command -- A command intended to be ran to completion (e.g. `ember build` or `ember test`).
+
+# Detailed design
+
+The following mechanism will be used to select the environment by each ember-cli command (in order):
+
+1. Command line specified environment (e.g. `ember serve --environment production`)
+2. ember-cli specific process environment variable specified environment (e.g. `EMBER_ENV=production ember serve`)
+3.  General purpose process environment variable specified environment (e.g. `NODE_ENV=production ember serve`)
+4. No environment was specified, default based on command type:
+	* Test commands -- Default to `test` environment.
+	* Interactive commands -- Default to `development` environment.
+	* Non-interactive commands -- Default to `production` environment.
+
+A pseudo-code based example:
+
+```javascript
+function isInteractiveCommand(commandName, knownArgs) {
+  switch (commandName) {
+  case 'test':
+    return knownArgs.includes('server');
+  case 'build':
+    return knownArgs.includes('watch');
+  case 'server':
+    return true;
+  default:
+    return false;
+  }
+}
+
+function determineEnvironment(commandName, args) {
+  let knownArgs = Object.keys(args);
+
+  // handle command line specified environment
+  if (knownArgs.includes('environment')) {
+    return args.environment;
+  }
+
+  // handle ember-cli specific process environment variable
+  if (process.env.EMBER_ENV !== undefined) {
+    return process.env.EMBER_ENV;
+  }
+
+  // handle general purpose process environment variable
+  if (process.env.NODE_ENV !== undefined) {
+    return process.env.NODE_ENV;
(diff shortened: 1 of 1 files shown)


hjdivad (approved):
I like this approach

active/0109-environment-selection.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Detailed design
+
+The following mechanism will be used to select the environment by each ember-cli command (in order):
+
+1. Command line specified environment (e.g. `ember serve --environment production`)
+2. ember-cli specific process environment variable specified environment (e.g. `EMBER_ENV=production ember serve`)
+3.  General purpose process environment variable specified environment (e.g. `NODE_ENV=production ember serve`)
+4. No environment was specified, default based on command type:
+	* Test commands -- Default to `test` environment.
```
trentmwillis: Based on the definitions above, "test commands" could also be considered "interactive" (`ember test --server`) or "non-interactive" (`ember test`), therefore I think we should specify an order of precedence here as well.

active/0109-environment-selection.md
```diff
@@ -0,0 +1,90 @@
...
+
+  if (commandName === 'test') {
+    return 'test';
+  }
+
+  return isInteractiveCommand(commandName, knownArgs);
+}
+```
+
+This detection will be done early in the command process, and the `process.env.EMBER_ENV` environment variable will be set with the selected environment (so that addons and other tools are able to know the environment being used).
```
trentmwillis, line 78: This may not be an issue, but having `EMBER_ENV` be the source of truth for which environment is being used seems potentially confusing when `EMBER_ENV` is not the highest specificity. Theoretically it would be possible to specify `EMBER_ENV` and a CLI flag in the same command and overwriting `EMBER_ENV` with the other value might cause confusion.

Again, may not be an issue, but something to consider.

trentmwillis (commented):
Overall a big :+1: on this. I know determining environment with addons has always been a pain point, so this would be fantastic.

active/0109-environment-selection.md
rwjblue: Ya, thats what I was trying to do by ordering these bullet points. Specifically, this is why `Test commands` is first.  I can redo it as an ordered list instead which would look like:

1. Command line specified environment (e.g. `ember serve --environment production`)
2. ember-cli specific process environment variable specified environment (e.g. `EMBER_ENV=production ember serve`)
3.  General purpose process environment variable specified environment (e.g. `NODE_ENV=production ember serve`)
4. No environment was specified, default based on command type:
	1. Test commands -- Default to `test` environment.
	2. Interactive commands -- Default to `development` environment.
	3. Non-interactive commands -- Default to `production` environment.

Does that make it better?

active/0109-environment-selection.md
hjdivad: I agree it's something to consider, but I think there's a lot of value in

1. Dealing with the various sources within ember-cli and giving addons a single place to look for the definitive environment
2. For that single definitive place to be something standard like a process ENV var.

rwjblue: I totally agree that this is confusing, but changing this would be somewhat breaking (since we already do this [here](https://github.com/ember-cli/ember-cli/blob/816b540e741345eeb54756c9f5c7a5304d6108da/lib/models/builder.js#L45-L46)).

active/0109-environment-selection.md
trentmwillis: Yep, I think just making it ordered should suffice.

active/0109-environment-selection.md
trentmwillis: I agree with @hjdivad, my concern is simply over potentially mutating a user-specified environment variable, but if it is already happening, then it seems likely to be a non-issue.

hjdivad: @trentmwillis i agree with the general concern around mutating user-specified env vars, but the only item that has precedence over `EMBER_ENV` in the rfc is the command line flag which means if the user specifies `EMBER_ENV` it will not be overwritten unless they *also* specify a command-line flag like `--environment production`

If a user does something like `EMBER_ENV=development ember build -prod` I don't think they'll think it too unreasonable for `EMBER_ENV` to change.

hjdivad: (also waiting for either you or rob to break the combo of everybody starting their replies by pointing out that they agree)

rwjblue:
We had a discussion about this in the ember-cli core team call today.  The team is generally 👍 to the long term goal, but would like a bit more massaging around implementing in a SemVer compatible way.  

Specifically, the suggestion is:

* Add a new flag to `.ember-cli` file (`defaultNonInteractiveCommandEnvironment`) which defaults to `production` for newly generated apps (and older apps that follow the standard `ember init` process).
* Detect if the flag is not present, and issue a warning to the console but fallback to existing behaviors (e.g. `development` by default).
* In a future version (possibly as early as 2.18.0 or as late as 3.0.0) remove the requirement for the flag, and use the scheme documented currently.

I'll work to update the RFC for that soon...

rtablada:
@rwjblue putting more information in the motivations will make this RFC easier for people to understand, read, and give feedback.

---

> We have received feedback from our direct users, developer relations folks, and benchmark maintainers suggesting that our current system for selecting the environment does not set our users up for success.

This is relevant for people close to core but outside of the news cycle and for those who aren't in the know on these events, this gives little to no context.
What feedback have we received?
What defaults do they recommend should be in place?
Why does having this new default mater?
More explanation of what this means would be helpful and there is not any mention of the current environment default, this is left to the user.

It would be helpful to give a little bit of background to the current environments and how they are used. Maybe something to the effect of:

```
Ember CLI currently allows users to specify the environment for any command using the flag `--environment`. This flag can be used by addons, build tooling, and the `config/environment.js` file to customize configuration, minification, and more.
```

---

I have a suggested alternative that we adjust and add things to the npm scripts:

The `build` script become `ember build -e production`.
Adding `-e test` to the test even though it is redundant.

From experience in other communities, they use the set of NPM scripts to teach people how to customize their own builds and show some common best practices.

The good thing about this alternative is that it would require no extra tooling and would have no breaking changes.
It also doesn't prevent the changes proposed in this RFC for future iterations.

rwjblue:
Closed as part of the efforts to consolidate RFC repos, will reopen in emberjs/rfcs...

PR #110, opened 2017-09-12
Packaging

twokul:
[Rendered](https://github.com/twokul/rfcs-1/blob/tree-shake-rfc/active/0051-packaging.md).

Related work:
+ [[INTERNAL] Introduce `broccoli-assembler`](https://github.com/ember-cli/ember-cli/pull/7317)
+ [Introduces Bundler](https://github.com/ember-cli/ember-cli/pull/7264)
+ [Extract vendor generation into bundler](https://github.com/ember-cli/ember-cli/pull/7269)
+ [[WIP] Treeshaking](https://github.com/ember-cli/ember-cli/pull/7308)


kellyselden:
With the `app.import` deprecation, does this RFC try to touch on or solve the problem if importing commonjs?

active/0051-strategies.md
```diff
@@ -0,0 +1,248 @@
...
+# Motivation
+
+The [Packager RFC](https://github.com/chadhietala/rfcs/blob/packager/active/0002-packager.md) submitted by [Chad Hietala](https://github.com/chadhietala) is a little over 2 years old. A lot of things have changed since then and it requires a revision.
+
+The current application build process merges and concatenates input broccoli trees. This behaviour is not well documented and is a tribal knowledge. While the simplicity of this approach is nice, it doesn't allow for extension. We can refactor our build process and provide more flexibility when desired.
+
+Most importantly, it helps us achieve:
+
++ defining and developing a common language around the subject
++ removing highly coupled code and streamline technical implementation (Ember Engines)
```
rwjblue: Fastboot is another example of this ([see here](https://github.com/ember-fastboot/ember-cli-fastboot/blob/c5ad2b7bd34ed28bbddebfdcc61bf555f8183e2c/index.js#L140-L174)).

active/0051-strategies.md
```diff
@@ -0,0 +1,248 @@
...
+
+The goal of this RFC is to solidify the next version of the mechanism that `ember-cli` uses to build final assets. It will allow for a more flexible build pipeline for Ember applications. It also unlocks building experimental features on top. It is a backward compatible change.
+
+# Motivation
+
+The [Packager RFC](https://github.com/chadhietala/rfcs/blob/packager/active/0002-packager.md) submitted by [Chad Hietala](https://github.com/chadhietala) is a little over 2 years old. A lot of things have changed since then and it requires a revision.
+
+The current application build process merges and concatenates input broccoli trees. This behaviour is not well documented and is a tribal knowledge. While the simplicity of this approach is nice, it doesn't allow for extension. We can refactor our build process and provide more flexibility when desired.
+
+Most importantly, it helps us achieve:
```
alexander-alvarez: The term `it` is vague in this context, should probably be `Strategies` or something else

active/0051-strategies.md
```diff
@@ -0,0 +1,248 @@
...
+
+# Unresolved questions
+
++ Will there be an ordering problem (strategies)?
++ Will it increase build time?
++ Which strategies are going to be supported out-of-the-box?
++ Should we introduce the same API on add-on level? 
++ Will simple extension strategy “just work”?
++ Communication between strategies
+
```
alexander-alvarez: How would we verify that a given strategy (or set of strategies) produces a valid application output?

rwjblue:
@kellyselden:

> With the app.import deprecation, does this RFC try to touch on or solve the problem if importing commonjs?

This RFC **is not** proposing a new deprecation. That mention is under a section titled "Topics for Future RFC's" which is only included to show a bit of future thinking/roadmap for the feature...

twokul:
@kellyselden just to expand on what @rwjblue already said. `app.import` deprecation requires an RFC on its own so it didn't make sense to go into details as it wasn't the purpose of this RFC.

kellyselden:
@twokul Sounds good. I misread the RFC.

active/0051-strategies.md
```diff
@@ -0,0 +1,248 @@
...
+## Assembler
+
+`Assembler` is similar to a car assembly line - starting with a metal frame, making changes to it along the way, adding parts and at the end, there's an assembled car.
+
+It does not make any assumptions about how the final output should be constructed. Instead, it relies on strategies to tell it what the final output should look like. This is rather important as it enforces clear separation of concerns.
+
+Assembler must have the following interface:
+
+```typescript
+interface Assembler {
```
simonihmig: Is this going to be a public API? So are there use cases for different `Assembler` implementations? Or is this just for clarification, but the end just an (internal) implementation detail?

active/0051-strategies.md
```diff
@@ -0,0 +1,248 @@
...
+## Strategies
+
+Strategy is an extensibility primitive. It gives you granular control over the final output. It could be used in many different ways (we are going to go over use cases below). 
+
+Strategy must have the following interface:
+
+```typescript
+interface Strategy {
+  constructor(options: any);
+  toTree(assember: Assembler, inputTree: BroccoliTree): BroccoliTree;
```
simonihmig: Curios: are there use cases where a `Strategy` needs access to the `Assembler`? Examples don't show this, and given the `Assembler` interface which only consists of constructor and `toTree`, looks like this could be omitted/further decoupled?

active/0051-strategies.md
twokul: Good question. We don't validate application output right now (usually build fails and that's how you know that something went wrong). We definitely need to provide an easy way to debug strategies which would help address some of the build issues. My assumption is that a "valid" output would depend on the application itself and suspect we need more experimentation on that front.

active/0051-strategies.md
twokul: `Assembler` is more of a concept, than anything else. If you start thinking in terms of strategies, it is nothing more but another type of strategy. [As of right now](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/assembler.js), it is an internal construct.

active/0051-strategies.md
twokul: Yes. Let's say we want to remove unused modules from an Ember application. `Assembler` could provide you with a list of modules/files that are used. That way it will be easy for a strategy to produce a broccoli tree with only relevant content. Some pseudocode (not a real implementation):

```javascript
const concat = require('broccoli-concat');

class DeadModulesEliminationStrategy() {
  constructor(options) {
    this.options = options;
  }

  toTree(assembler, inputTree) {
    return concat(inputTree, {
      // ... code ...
      inputFiles: assembler.getUsedModules()
      // ... code ...
    };
  }
}
```

active/0051-strategies.md
```diff
@@ -0,0 +1,247 @@
...
+}
+```
+
+`inputTree` will have the following structure:
+
+```sh
+bundler:js:input/
+├── addon-modules
```
mixonic: This is the directory name? Is is expected that this string would be split or otherwise processed?

(Also, this is not `sh` syntax afaict)

active/0051-strategies.md
```diff
@@ -0,0 +1,247 @@
...
+└── vendor
+```
+
+where `addon-modules` contains dependencies from Ember add-ons; `the-app-name-folder` contains Ember application code; `node_modules` contains node dependencies and `vendor` contains other dependencies.
+
+Note, that  `toTree` method must return a broccoli tree.
+
+### Concatenation Strategy
+
+One of the strategies that we already use now to produce `assets/app-name.js` and `assets/vendor.js`. It takes a broccoli tree and returns a new concatenated broccoli tree.
```
mixonic: First sentence here needs to be restructured. You might want to make it clear that although you are proposing Strategies as a public API, a similar architecture is already used internally in Ember-CLI. I think that is implied in your statement.

active/0051-strategies.md
```diff
@@ -0,0 +1,247 @@
...
+
+Where `this.options` mimics [`broccoli-concat` options](https://github.com/broccolijs/broccoli-concat#advanced-usage).
+
+### Example of possible strategies
+
+Below are some of the strategies that could be implemented.
+
+#### Assets Split Strategy
+
+Other strategy that proved very useful is an ability to separate assets. One of the applications would be to separate "static" assets from application code. They are called "static" because they don't change very often and could stay cached by CDN throughout several deployments (`ember.js`, `jQuery` or third party dependencies). That way performance is less effected by deployments.
```
mixonic: First sentence needs to be restructured. I think this generally can be clearer.

mixonic: This example is not just an example, it is how bundling works today. Are you suggesting that it *could* be ported to a strategy or that it *be* ported?

Other items in the "examples" section here seem to be only for illustration, but this one I would expect will be written as a built-in.

active/0051-strategies.md
```diff
@@ -0,0 +1,247 @@
...
+  const secondStrategy = new MyCustomStrategy();
+  const stragies = defaultStrategies.concat(firstStrategy, secondStrategy);
+
+  const app = new EmberApp(defaults, { strategies });
+
+  return app.toTree();
+}
+```
+
+Note, that `strategies` is an optional property. If you don't pass it to the constructor or `strategies` is an empty array, you effectively "opting out" of the feature and using the default behaviour.
```
mixonic: you => you're

active/0051-strategies.md
```diff
@@ -0,0 +1,247 @@
...
+  as improve repeat page load performance)
+
+# How We Teach This
+
+This is a backward compatible change to the existing Ember CLI eco system. In order to teach users how to use strategies, we need to update the API docs with a section for this and the best practices of when to use this. A more general purpose blog post could be beneficial as well.
+
+# Drawbacks
+
+- More libraries to maintain
+- Increasing public API surface
```
mixonic: I encourage you to think through this section more instead of only providing it lip service. Some ideas to start with:

* Does this feature further couple Ember apps and addons to Broccoli? Is that acceptable or a risk we can mitigate?
* `defaultStrategies` is an array in your example above. Is it ok that someone could `push` onto that array? Should we consider freezing the object or instead exporting a function that generates a new array on each invocation?
* Are there existing addons dependent on the current bundling implementation in Ember-CLI?
* Are there build performance concerns?
* Are the files being bundled only JavaScript, for example post HBS compilation? How can we bundle intelligently when we have a mix of HBS and JS files, or is that concern handled elsewhere?

I see you have some of these under unresolved, maybe some eventually will become drawbacks.

active/0051-strategies.md
```diff
@@ -0,0 +1,247 @@
...
+
+# Alternatives
+
+[Webpack](https://webpack.js.org) has become very popular in solving this similar problem. However it is unclear how WebPack could be used with Ember as Ember has both explicit and implicit dependencies. This is why we feel we need something that is aware of how Ember apps are assembled and take advantage of existing tools when it is makes sense. Since we have an opinionated way of structuring our apps, we can avoid the "wall of configuration" that other asset packaging systems are susceptible to.
```
mixonic: > it is unclear how WebPack could be used with Ember as Ember has both explicit and implicit dependencies.

Can you provide examples?

> we feel we need something that is aware of how Ember apps are assembled

You've gone into some effort in this RFC to make clear that strategies are agnostic bundling steps. I don't see how they are coupled to how Ember apps are assembed, can you provide an example?

active/0051-strategies.md
twokul: I suspect I need to rewrite this section to be more clear as you suggested. Current bundling works differently (unless I misunderstood what you meant) that is _all_ third party libraries end up in one `vendor.js` file as opposed to being bundled separately.

I think this strategy could be provided out of the box by Ember CLI.

active/0051-strategies.md
twokul: all valid points and @rwjblue and I discussed all of them. having said that, we definitely need to have them in writing. let me address that.

fteotini:
> This behaviour is not well documented and is a tribal knowledge.

This!
I always thought that broccoli build system is simple and powerful but it's nearly impossible to find any decent advanced documentation on how it works and how extend it. So any change towards better extension and ease of understanding is well accepted.
I'll read more carefully the RFC and try to provide useful comments XD

dherman:
Something I couldn't quite figure out from reading the RFC: does this API provide a mechanism for hooking into the built-in ember-cli build pipeline, or only to replace it entirely? In particular what I'm interested in is being able to hook in to the part of the default pipeline where modules have been parsed, and access their ASTs. If plugins have to replace the pipeline entirely then if they need semantic information from ASTs they'd have to parse the source files themselves, resulting in everything in the whole directory tree being parsed twice.

twokul:
@dherman thanks for reading!

> does this API provide a mechanism for hooking into the built-in ember-cli build pipeline, or only to replace it entirely?

As it stands right now, it merely introduces a way for developers to have granular control over bundling. At the moment, developers have no say in what goes where, how many bundles get produced, what to drop and what to include.

By introducing `strategies`, we allow for this flexibility and don't break the current behaviour of the build pipeline.

> In particular what I'm interested in is being able to hook in to the part of the default pipeline where modules have been parsed, and access their ASTs. If plugins have to replace the pipeline entirely then if they need semantic information from ASTs they'd have to parse the source files themselves, resulting in everything in the whole directory tree being parsed twice.

This is something @rwjblue and I discussed as well but I decided to not include it in this RFC as it would create more questions than provide answers.

In order for us to incorporate fully functional treeshaking into the build pipeline, we absolutely need ASTs and don't want to parse files twice. The tricky part here is to make sure we expose the right public API w/o shooting ourselves in the leg. This API requires not only more experimentation but changes to the way the build pipeline works today. 

Having said all of that, we want to make this API a reality and could definitely benefit from real life use cases like the one you mentioned.

twokul:
@dherman to expand on my previous answer a little. Use case that you mentioned was one of the main drivers behind passing an instance of `Assembler` to `toTree` function. We could iterate on the specific APIs within `Assembler` that will allow to access ASTs.

luxzeitlos:
Does the assembler call the strategies in parallel or in seriel? If it's run in parallel, how to modify the final output, and if it's run in seriel, wouldn't need the strategy to provide the same directory structure as output as specified as input for strategies? And how is that different from the abstraction we already have on broccoli layer, which maps input trees to output trees? Isn't that exactly what a structure does?

twokul:
@luxferresum 

> Does the assembler call the strategies in parallel or in seriel?

It doesn't run in parallel. It might be rather challenging to coordinate that b/c Broccoli itself does not have such primitives and I don't think there are any plans to support that.

> wouldn't need the strategy to provide the same directory structure as output as specified as input for strategies?

Not really, strategies get access to the same input tree, they are not chained off of each other.

> And how is that different from the abstraction we already have on broccoli layer, which maps input trees to output trees?

I'm not too sure what abstraction you're referring to here, I'm going to assume it's the way we use Broccoli within Ember CLI. If so, the idea behind strategies is to make it easier to the way input trees are mapped to output trees and hence have control over the final bundles.

Let me know if I misunderstood your questions!

luxzeitlos:
Thanks for your answer!

> strategies get access to the same input tree, they are not chained off of each other.

Thats what I mean with running "in parallel". Sorry for being unclear, I basically wanted to ask if they are chained off on each other or run independent, and the final output of all strategies is merged in the end. So thanks for clarifying this. 

However this keeps my subsequent question unanswered:

> how to modify the final output

If the strategies are applied independent of each other, there is no way to modify the output of a "default" strategy and doing any kind of post-processing. Is this correct? Same is true for preprocessing.

Also this is very confusing for me because it does not match the idea of a car assembly line:

> `Assembler` is similar to a car assembly line - starting with a metal frame, making changes to it along the way, adding parts and at the end, there's an assembled car.

This suggest that the car is build step by step, starting from the metal frame, making changes on the way, step by step, each step using the output of the previous step and adding something until I have a car. So This suggests the same for the ember `Assembler`, that it would run the first strategy, and then all others, each depending on the previous one, building an an assembly line from the development directory tree to the final build output.

So either the comparison of the `Assembler` and a assembly line is very wrong, or I'm missing something here?

Let me try to clarify. Either you run the strategies in subsequent series:


```
->strategyA->strategyB->strategyC->
```

Or independent:

```
    __> strategyA ___
   /                  \
__/___> strategyB _____\__>
  \                    /
   \__> strategyC ____/

```

If you run them independent, how to do post processing and how is this an assembly line?

---

my reference to the abstraction layer is the Broccoli Plugin Layer itself. If I understand a `strategy` basically maps one `inputTree` to one `outputTree`! And thats the same that a broccoli plugin does (simplified):

```
treeWithCss = compileSass(treeWithSass)
```

Then how is strategy different from a broccoli plugin? I'm expecting most strategies to be just wrappers around a broccoli plugin, or maybe a series of broccoli plugins.

twokul:
@luxferresum sorry for the delayed response and thanks for explaining. I can see how it could be confusing now 😕 

> Also this is very confusing for me because it does not match the idea of a car assembly line:

They are indeed different. If we go along with a car-related analogy here, it's more of a service shop where the whole car gets "decorated" (repaint, rims/wheels change, etc.) 🚗  Let me change that part and thanks for pointing it out!

> Then how is strategy different from a broccoli plugin? I'm expecting most strategies to be just wrappers around a broccoli plugin, or maybe a series of broccoli plugins.

The underlying primitive is indeed a broccoli plugin (it has nothing to do with `strategies`, that's how broccoli works). And yes, strategies will likely end up being wrappers around broccoli plugins. Going down the "everything is a broccoli plugin" path would lead to even further coupling of Ember CLI and Broccoli which actually doesn't have to be so. `Strategy` is more of a build pipeline primitive that can hide the build tool completely; "register a transform and it will happen" kind of thing. That way we could theoretically abstract "common" parts of CLI and make it less ember specific but keep interfaces between components in place.

Let me know if my answers made sense and let me rephrase parts of the RFC to include your feedback! 💪

luxzeitlos:
Thats for your answer! Well, the one thing I still don't understand is how I would do something like a postprocessing with an strategy? Or is this out of scope?

twokul:
> Well, the one thing I still don't understand is how I would do something like a postprocessing with an strategy?

`strategies` aren't meant to be used for postprocess transformations; they are just solutions to different problems. The mental model here would be: `postprocessTree` is called _after_ `strategies` have been processed.

> Or is this out of scope?

at this point it is but I think it's good to disambiguate in the RFC

active/0051-strategies.md
```diff
@@ -0,0 +1,368 @@
...
+For example:
+
+```javascript
+// ember-cli-build.js
+const { defaultStrategies } = require('ember-cli/lib/strategies');
+
+module.exports = function(defaults) {
+  const firstStrategy = new MyCustomStrategy();
+  const secondStrategy = new MyCustomStrategy();
+  const stragies = defaultStrategies.concat(firstStrategy, secondStrategy);
```
trabus: Silly nitpick, but thought I should point out the typo here: `stragies`

active/0051-strategies.md
```diff
@@ -0,0 +1,368 @@
...
+Ember applications would be able to register custom strategies via
+`ember-cli-build.js` file by passing a `strategies` property to the constructor.
+
+Passing `strategies` as an option to the constructor will clobber default
+strategies that Ember CLI provides out of the box (they mimic existing behaviour
+of creating `dist` folder with final assets).
+
+However, Ember CLI will expose an array of default strategies that people can
+take advantage of. Note, that `defaultStrategies` will be freezed so you can't
+push directly onto it. One would need to create a new array.
```
trabus: I'm assuming the processing of these is simply in they order they appear in the array, though I was curious if there would be need for some sort of priority system to allow some finer control over order of operations in case it was needed.

trabus: I would assume that priority system would be implemented in the strategies themselves and would be managed by the assembler when processing the strategies (sort of like how addons are handled currently)?

twokul: I mention this in "Unresolved Question" section. There might be an ordering problem but before coming up with a solution, I think we should surface a real use case first. Add-ons don't have a priority system and things "just work".

> I would assume that priority system would be implemented in the strategies themselves 

That's one way to implement it, I would stay clear for a particular implementation until we know more about constraints.

trabus: Actually, in the addon config (inside the `package.json`), you can define an array of `before` and `after` to ensure that addons are resolved in the desired order, which then is managed by the dependency graph:
https://github.com/ember-cli/ember-cli/blob/master/lib/models/addons-factory.js#L41

We use this strategy for `ember-cli-legacy-blueprints`, to ensure that if `ember-core` is installed, its blueprints will win:
https://github.com/ember-cli/ember-cli-legacy-blueprints/blob/master/package.json#L55

active/0051-strategies.md
hjdivad: @twokul it seems odd to add a parameter solely as a bag of potential future API.  I had the same question as @simonihmig when reviewing.

active/0051-strategies.md
```diff
@@ -0,0 +1,368 @@
...
+constructed. Instead, it relies on strategies to tell it what the final output
+should look like. This is rather important as it enforces clear separation of
+concerns.
+
+`toTree` method would gather all the trees from the passed in `strategies`, and
+merge them into one final broccoli tree.
+
+## Strategies
+
+Strategy is an extensibility primitive. It gives you granular control over the
```
kratiahuja: I think it would be nice to provide guidance on how this differs from the similarly existing `app.import` with `using` API.

active/0051-strategies.md
```diff
@@ -0,0 +1,368 @@
...
+certain type. In fact, it's very similar to `ConcatenationStrategy` described
+above.
+
++ reduces the API surface (as opposed to `ember-engines` and
+  `ember-asset-loader`, only one strategy)
++ makes things explicit, meaning that you have to register a strategy
+
+## Registering custom strategies
+
+Ember applications would be able to register custom strategies via
```
kratiahuja: What about Ember addons? I believe a lot of addons like fastboot may want to provide their own strategies that should be registered as well.

active/0051-strategies.md
```diff
@@ -0,0 +1,368 @@
...
+concerns.
+
+`toTree` method would gather all the trees from the passed in `strategies`, and
+merge them into one final broccoli tree.
+
+## Strategies
+
+Strategy is an extensibility primitive. It gives you granular control over the
+final output. It could be used in many different ways (we are going to go over
+use cases below). Note, that aren't meant to be used for "postprocess"
```
kellyselden: that -> they

active/0051-strategies.md
```diff
@@ -0,0 +1,368 @@
...
+
+Ember applications would be able to register custom strategies via
+`ember-cli-build.js` file by passing a `strategies` property to the constructor.
+
+Passing `strategies` as an option to the constructor will clobber default
+strategies that Ember CLI provides out of the box (they mimic existing behaviour
+of creating `dist` folder with final assets).
+
+However, Ember CLI will expose an array of default strategies that people can
+take advantage of. Note, that `defaultStrategies` will be freezed so you can't
```
kellyselden: freezed -> frozen

active/0051-strategies.md
```diff
@@ -0,0 +1,368 @@
...
+push directly onto it. One would need to create a new array.
+
+For example:
+
+```javascript
+// ember-cli-build.js
+const { defaultStrategies } = require('ember-cli/lib/strategies');
```
kellyselden: Is it a good idea to have people reach in like this? I could imagine `EmberApp.defaultStrategies` being a more user-friendly way of getting them.

active/0051-strategies.md
```diff
@@ -0,0 +1,368 @@
...
+in Lisp. The gist of the idea is "how about we start using _only_ the code that
+we actually need?"
+
+Secondly, how is it different from [Dead Code
+Elimination](https://en.wikipedia.org/wiki/Dead_code_elimination)? [Rich
+Harris](https://twitter.com/Rich_Harris)
+[offers](https://medium.com/@Rich_Harris/tree-shaking-versus-dead-code-elimination-d3765df85c80)
+a pretty good explanation in the context of [Rollup](https://rollupjs.org/). The
+gist is dead code elimination happens on a final product by removing bits that
+are unused. Treeshaking is not about excluding the code but about including the
```
kellyselden: "Treeshaking is not about excluding the code but about including the code that's used."
This line is super confusing to me and actually makes it harder for me to understand the difference. It could be just me though.

active/0051-strategies.md
```diff
@@ -0,0 +1,368 @@
...
+
+# Drawbacks
+
+There are several potential drawbacks that are worth noting.
+
+_Build performance_. There is minimal overhead in instantiating strategies and
+calling methods on them and I believe this approach shouldn't degrade build
+performance.
+
+_Coupling with Broccoli_. The idea of having a more generic cli was discussed
```
kellyselden: This is probably not worth mentioning IMO. The discussions are making it generic were Ember CLI w/ broccoli + angular/react/vue etc. This RFC wouldn't affect any of those considerations.

kellyselden:
I just remembered the crazy component creation I'm doing here https://github.com/ember-vr/ember-aframe/blob/master/index.js#L104-L117. @twokul Would this RFC affect that code, would it become invalid?

active/0051-strategies.md
```diff
@@ -0,0 +1,368 @@
...
+  }
+}
+```
+
+#### Debug/Analyse Strategy
+
+Yet another application of strategies would be to run different analysis on the
+Ember applications. Take
+[broccoli-concat-analyser](https://github.com/stefanpenner/broccoli-concat-analyser),
+for example. This could be a strategy and easily encorporated into the build.
```
sangm: encorporated -> incorporated

active/0051-strategies.md
```diff
@@ -0,0 +1,368 @@
...
+
+#### Engine Strategy
+
+Introducing `Strategy` as a concept allows us to re-think the way we approach
+Ember Engines as well. We can start thinking about engines as a strategy of
+certain type. In fact, it's very similar to `ConcatenationStrategy` described
+above.
+
++ reduces the API surface (as opposed to `ember-engines` and
+  `ember-asset-loader`, only one strategy)
```
sangm: How would this affect `ember-asset-loader`?

twokul:
> I just remembered the crazy component creation I'm doing here https://github.com/ember-vr/ember-aframe/blob/master/index.js#L104-L117. @twokul Would this RFC affect that code, would it become invalid?

It shouldn't break your code, no. Tested it with a bunch of different add-ons like `ember-cli-mirage` and `ember-asset-loader` @kellyselden

active/0051-strategies.md
twokul: In short term, `ember-asset-loader` is not going away but I can imagine it changing API surface a little (once lazy loading w/o engines becomes possible)

active/0051-strategies.md
```diff
@@ -0,0 +1,441 @@
...
+const FastbootStrategy = require('ember-fastboot/strategies/fastboot');
+
+module.exports = function(defaults) {
+  const app = new EmberApp(defaults, { });
+
+  app.strategies = [
+    new FastbootStrategy(),
+    createDefaultApplicationStrategy(app),
+    ...createDefaultVendorStrategy(app)
+
```
kellyselden: node 4 code here too

active/0051-strategies.md
```diff
@@ -0,0 +1,441 @@
...
+  }
+}
+
+module.exports = function(defaults) {
+  const app = new EmberApp(defaults, { });
+
+  app.strategies = [
+    new Noop(),
+    createDefaultApplicationStrategy(app)
+  ].concat(createDefaultVendorStrategy(app));
```
kellyselden: Why does `createDefaultVendorStrategy` return an array of strategies?

twokul: it's mostly have to do with the fact that vendor can produce several files, not just one.

active/0051-packaging.md
```diff
@@ -0,0 +1,338 @@
...
+  + optimising plugins (JavaScript and CSS tree-shaking)
+
+# Scope
+
++ New public API for customising build process and giving more granular control over
+the final build output
+
+# Terminology
+
++ **Packaging** - A process during of designing, evaluating, and producing final build assets.
```
kellyselden: "The process of" might be what you mean?

active/0051-packaging.md
```diff
@@ -0,0 +1,338 @@
...
+future RFCs, but the discussion points have been included below.
+
+## Tree-shaking
+
+Firstly, what's _tree-shaking_? AFAIK, the term
+[originated](https://groups.google.com/forum/#!msg/comp.lang.lisp/6zpZsWFFW18/-z_8hHRAIf4J)
+in Lisp. The gist of the idea is "how about we start using _only_ the code that
+we actually need?"
+
+Secondly, how is it different from [Dead Code
```
kellyselden, line 223: You can similarly explain this as dead code elimination is an imperfect outside-in approach, and tree-shaking is a (near) perfect inside-out approach.

active/0051-packaging.md
```diff
@@ -0,0 +1,338 @@
...
+similar problem. One could implement a `package` function that would use Webpack
+for packaging. Ultimately, we need something that is aware of how Ember apps are
+assembled and how Ember apps utilise dependency injection that takes advantage
+of existing tools. The long term plan is to have a dependency graph that is
+aware of application structure so can avoid the "wall of configuration" that
+other asset packaging systems are susceptible to.
+
+# Unresolved questions
+
++ Will it increase build time?
```
kellyselden, line 328: I don't think we will be doing any extra work.

active/0051-packaging.md
```diff
@@ -0,0 +1,338 @@
...
++ Should we introduce the same API on add-on level?
```
kellyselden, line 329: I don't think so, since the end goal is for the app to compile addon code during it's tree-shaking step.

active/0051-packaging.md
```diff
@@ -0,0 +1,338 @@
...
+still be able to customise the final build output.
+
+```javascript
+// ember-cli-build.js
+const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+const packagerFor = require('ember-cli/lib/broccoli/packager-for');
+
+module.exports = function(defaults) {
+  const app = new EmberApp(defaults, {
+    package(inputTree) {
```
kellyselden: I still think we should still explore breaking this into two: `transform` and `package`. Then we could eliminate the `packageFor` and `defaultPackager` API. If someone needs to tree-shake and not mess with the packaging, they would use `transform`. If they need to modify the packaging, they weren't going to use the defaults anyways so they could just use `package`.

active/0051-packaging.md
```diff
@@ -0,0 +1,338 @@
...
+      const defaultPackager = packagerFor(app);
+      // customise `inputTree`
+
+      return defaultPackager(inputTree);
```
kellyselden: Assuming the `package` function receives a precompiled tree, if someone wanted to do custom bundling for their app, would they be opting out of our compilation system? Would they be responsible for doing their own compilation, even though they didn't want to affect the compilation step?

active/0051-packaging.md
```diff
@@ -0,0 +1,338 @@
...
+the reasoning is very simple: if `ember.js` or `jQuery` (possibly along with
+other third party libraries) don't change with every deployment, why bust CDN
+cache for them?
+
+#### ES6 Modules
+
+ES6 modules are [starting](https://caniuse.com/#feat=es6-module) to land in browsers.
+This means that you can use `<script type="module" src="/my/app.js"></script>`.
+
+This [article](https://philipwalton.com/articles/deploying-es2015-code-in-production-today/) [explains](https://philipwalton.com/articles/deploying-es2015-code-in-production-today/#is-this-really-worth-the-extra-effort) the benefits of using ES6 modules over ES2015 (smaller total file sizes, faster to parse and evaluate).
```
Turbo87, line 205: > ES6 modules over ES2015

🤔

active/0051-packaging.md
twokul: The idea was to introduce as few new APIs as possible. I very much agree with idea of separating side effects (for example treeshaking) from packaging and I’m not opposed to “transform”. However it sounds very generic. Also, the long term goal here is to make treeshaking “hidden” from the developers perspective exposing only either some form of configuration or dependency graph itself.

If understood you correctly, the idea is to swap “packagerFor” with “transform”? Do you see any downsides of introducing “packagerFor” API?

active/0051-packaging.md
twokul: Generally, yes. I was toying around with the idea of maybe exposing some of the dep graph APIs on the add-on level. Having said that, it might be too far in the future

active/0051-packaging.md
```diff
@@ -0,0 +1,338 @@
...
+├── node_modules/
+└── vendor/
+```
+
+where:
+
++ `addon-tree-output` is a folder that contains dependencies from Ember add-ons
++ `the-app-name-folder` is a folder that contains Ember application code
++ `node_modules` is a folder that contains node dependencies
++ `vendor` is a folder that contains other dependencies.
```
sangm: Have we thought about including the test tree here?

We may want to produce separate bundles for lazy loading tests between an engine and the host

active/0051-packaging.md
```diff
@@ -0,0 +1,336 @@
...
+It's important to make it easy for users to still use default Ember CLI
+packaging.
+
+`defaultPackager` is a way for the users to access out-of-the-box packaging while
+still be able to customise the final build output.
+
+```javascript
+// ember-cli-build.js
+const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+const defaultPackager = require('ember-cli-default-packager');
```
ro0gr, line 142: Does it mean ember-cli-default-packager has to be listed in the project dependencies?

active/0051-packaging.md
```diff
@@ -0,0 +1,336 @@
...
+finished.
+
+Currently, Ember.js application and all of its depedencies get assembled under one
+directory with the following structure:
+
+```ruby
+bundler:js:input/
+├── addon-tree-output/
+├── the-app-name-folder/
+├── node_modules/
```
sangm, line 64: you need to include `tests` here in the diagram

twokul: it's the current directory structure, not the new one

rwjblue:
Moving to FCP...

devinrhode2:
Looks like the Rendered link at the top is broken :(

btecu:
You can just click **Files changed** and then **View**:
https://github.com/twokul/rfcs-1/blob/d26cb1ff75f78c83fcbabe7de57f365a970f120f/active/0051-packaging.md

buschtoens:
I've been starting to experiment with the API have one small complaint regarding ergonomics for the end user: It appears to not be possible for an addon to hook into the `package` hook. Users have to explicitly `require` and pass the packager to the `EmberApp`.

*EDIT:* I lied. 😂 You can access / override `parent.options.package` in the `included` hook just fine.

```js
module.exports = {
  name: require('./package').name,

  included(parent) {
    // Assuming this addon is directly included by the host app.
    // Otherwise you need to traverse `parent` upwards.
    parent.options.package = function(inputTree) {
      // Do what you want with `inputTree` here or just forward to the default packager implementation.
      // https://github.com/ember-cli/ember-cli/blob/v3.6.0-beta.1/lib/broccoli/ember-app.js#L1765-L1767
      return this._legacyPackage(inputTree);
    };

    this._super.included.apply(this, arguments); 
  }
}

PR #114, opened 2018-01-04
Add ember-cli-template-lint to default app & addon blueprints

knownasilya:
[Rendered](https://github.com/knownasilya/rfcs-1/blob/5fa645816567f4aeb66a97af0ae19d8f830212dc/active/0000-add-template-lint-addon.md)


kellyselden:
We would love to do this. One thing we would like to knock out before adding it to the blueprint is make it more official.

 * move to ember-template-lint org, or even ember-cli
 * add more admins, so all the maintenance work doesn't fall on Rob's back

Perhaps you could add these to the RFC?

knownasilya:
Updated

knownasilya:
@kellyselden is there anything stopping this from going into FCP?

knownasilya:
Woot!

PR #121, opened 2018-08-14
[RFC] Replace ember-cli-eslint with the standard eslint

sangm:
[RENDERED](https://github.com/sangm/rfcs/blob/master/active/0000-remove-ember-cli-eslint.md)

Initial draft


active/0000-remove-ember-cli-eslint.md
```diff
@@ -0,0 +1,51 @@
+- Start Date: 2018-08-13
+- RFC PR: (leave this empty)
+
+# Summary
+
+Remove https://github.com/ember-cli/broccoli-lint-eslint +
+https://github.com/ember-cli/ember-cli-eslint as a dependency within
+`ember-cli`.
```
dfreeman, line 7: Are you proposing the removal of lint hooks completely from ember-cli, or just removing `ember-cli-eslint` as a default dependency of new projects?

sangm: As far as I understand, only reason why we need the lint hooks is to trigger logic within `ember-cli-eslint`. (I assumed they were tightly coupled since they were marked as private).

If that's not the case, we should leave the link hooks as is.

dfreeman: There are some hooks beyond the ones you mentioned here, like [`lintTree`](https://ember-cli.com/api/classes/Addon.html#method_lintTree) that are used by [a number of addons in the ecosystem](https://emberobserver.com/code-search?codeQuery=lintTree%28).

But more broadly, `broccoli-lint-eslint` and `ember-cli-eslint` aren't dependencies of `ember-cli` itself, only of projects that the CLI generates. Given the phrasing here, I'm still a little unclear about whether you're proposing that Ember CLI stop supporting linters as part of the build, or just changing the default for new projects while leaving the ability for folks to opt in 🙂

sangm: I can see how that's unclear. I'll make some changes 👍 Thanks for the feedback @dfreeman

jnfingerle:
Thanks for taking your time to write this RFC, nonetheless I'm against this. 

Linting as part of the test is a big feature to enforce project standards. Editor integration and pre-commit hooks are great for developer convenience, but cannot replace this. 

pre-receive hooks could replace this, but (a) aren't available on hosted (standard) github and (b) would mean that everybody has to do something by hand that ember could have done for them.

I do understand, that I could just add ember-cli-eslint by hand and I would do this for each and every project I'd be creating. But the default blueprint is supposed to be a good starting point, not a minimal one.

sandstrom:
"No console warnings during builds" is a big drawback.

Will there be another, easy, way of getting them back if this RFC would be accepted?

joukevandermaas:
I agree with @jnfingerle. I don't think the motivations listed are a good trade-off here for most projects. I think that "most projects do/don't want this" should be the *main* motivation for adding or removing things from the default template. I think most javascript developers would agree that linters are invaluable, and *the best* time to add a linter is at the start of a project.

That said, perhaps some more detailed explanation of the motivations could change my mind on the trade-offs involved. I have to admit that I don't understand the context for point 2 and 3 at all, and point 1 is not backed up with data.

joshsearles:
I would be totally for this as an _Optional_  feature. Having linting out of the box available I believe enforces a common standard and is a big help to users who don't use an IDE to code or even a second layer precaution to those new to Ember.

active/0000-remove-ember-cli-eslint.md
```diff
@@ -0,0 +1,40 @@
...
+3. Simplicity. `eslint` is common among JS stack, and integrations with editors
+   / precommit-hooks are ubiquitous. Removing this layer of abstraction will
+   simplify how `eslint` is used throughout `ember-cli`.
+
+# Detailed design
+
+1. Change blueprint to pull in `eslint` as opposed to `ember-cli-eslint` under
+   `devDependencies`.
+2. Provide documentation on `eslint` and editor integration as well as precommit hooks
+
```
jrjohnson, line 41: Maybe:
3. Add to the default Travis blueprint:

```
script:
  - yarn lint:js
  - yarn test
```

To make it clear that linting is still running (and still part of the test process) it just doesn't run along side the other tests.

sangm:
@kellyselden / @rwjblue have deeper insights into how this decision may affect ember-cli: https://github.com/ember-cli/ember-cli/issues/7975

sangm:
As an example of how this affects ember-cli:

https://github.com/ember-cli/ember-cli-eslint/issues/56#issuecomment-407365145

happycollision:
Anecdotally, I always check the box to ignore lint errors when I am running tests in the browser because I am in “make it work” mode. When I am refactoring I check for lint errors. But I appreciate having those warnings (and some explanations) in my IDE during all phases of writing my code.

So I guess I am saying that if removing `ember-cli-eslint` would still allow ember-specific lint messages in my IDE, I am fine with that. But, as a beginner, I am very happy that those messages appeared out of the box.

active/0000-remove-ember-cli-eslint.md
sangm: I think this is a good idea. Thanks @jrjohnson!

jrjohnson:
@happycollision yes. I think that is missing from this RFC - that highlighting (at least in my IDE VS Code) is done by the standard eslint plugin.  If I understand it correctly nothing Ember is required. I think this RFC should make it clear that what `ember-cli-eslint` does is to lint during test runs. 

It is `Eslint` and your apps `.eslint.rc` file that do the actual linting and those will remain.

sangm:
Updated @jrjohnson

jnfingerle:
1. I don't see what place a consensus in the core team might have in this RFC. If there is already a consensus, they should just do it, so we don't have to use up our time on this RFC. If, on the other hand (and as I understood), the core team is interested in the community's view and weighs in the arguments in a RFC discussion, it's premature to talk about a consensus and at best an argumentum ad verecundiam (argument from authority). The RFC should rely on it's own merits, not on real or supposed approval from a special subset of the community. That said, I'm aware that this might still be the consensus in the end. But having it written into the RFC before the formal decision is made is leaving more than a bitter taste on my end.

2. "Drawback" is still missing, that `ember-cli-eslint` is the easiest way to have automated builds fail on eslint errors and warning, regardless of the tool used for building. 

Obviously you might have different views on the matter if a failed lint should be treated the same as a failed test. Since eslint contains many rules that aren't only about appearance but to enforce coding habits that are less error prone I more than tend to that it should be treated equally.

sangm:
@jnfingerle (1) is a good point. I'll take it out

buschtoens:
When this RFC was originally submitted I leaned towards 👎 , but after reading through https://github.com/ember-cli/ember-cli-eslint/issues/235, this changed to a strong 👍 .

As long as linting is still [part of the default CI, as requested here](https://github.com/ember-cli/rfcs/pull/121#discussion_r210662129), I am very much in favor of this. Providing an optional _good and fast_ default pre-commit / pre-push hook would also be super sweet!

active/0000-remove-ember-cli-eslint.md
```diff
@@ -0,0 +1,47 @@
...
+path, this should become _easier_ to teach because of less abstractions.
+
+# Drawbacks
+
+1. No console warnings during builds
+2. lint failures are no longer included in browser tests
+
+# Alternatives
+
+N/A
```
rwjblue: We at least mention the alternative of: "do nothing" (not that I think that is a good thing to do...)

active/0000-remove-ember-cli-eslint.md
rwjblue: @jrjohnson - You are totally correct! So much so, that its **already what we do!**

https://github.com/ember-cli/ember-new-output/blob/v3.4.1/.travis.yml#L25-L27

rwjblue (approved):
I completely agree with this RFC and think it is without a doubt what we should do going forward.

See https://github.com/ember-cli/ember-cli-eslint/issues/235 for yet another example of the issues that we face if we try to "stay in the middle".

rwjblue:
@jnfingerle 

> Linting as part of the test is a big feature to enforce project standards. 

I totally agree that linting is super important, and that projects should enforce linting standards before merging any changes. This RFC doesn't make any changes to ember-cli's stance on that, it doesn't make it more difficult to see linting failures (in fact with `yarn lint:js` / `yarn lint:hbs` I think it is actually _much_ easier to reason about), and it makes the Ember ecosystem "just another consumer" of standard JS tooling. 

@sandstrom 

> "No console warnings during builds" is a big drawback.

I agree, @sangm can you add that to the list of drawbacks? FWIW, I do think there may be other ways to solve this, and I also think that even with these changes we still get direct "in your face" feedback that you have linting failures (right there in your editor without having to flip around to the terminal)...

@happycollision 
 
> Anecdotally, I always check the box to ignore lint errors when I am running tests in the browser because I am in “make it work” mode. When I am refactoring I check for lint errors. But I appreciate having those warnings (and some explanations) in my IDE during all phases of writing my code.

Exactly! I also do exactly the same thing (disabling linting while running QUnit tests). When I'm ready to commit things, I either run `yarn lint:js` or leverage precommit hooks (though I am much less in favor of precommit hooks 😛). The key here is that we aren't talking "removing linting" and the feedback loop is really best when directly in editor anyways...

IMHO, this is absolutely important and we should move forward...

_Note: I happened to do my initial review of this RFC during an “issue triage” stream I was doing, you can check that out [here](https://www.twitch.tv/videos/307153695)_

---

I look forward to hearing folks comments / thoughts, please keep the conversation going... 😸

webark:
In a similar vein to adding a lint command to travis, for those that want the linting every time tests are run, how hard is it to add a `yarn test` or equivalent to the testem config to fail the tests when they are run as normal?

webark:
Also, i’d suggest changing the name to something more like “replace ember-cli-eslint with the standard eslint package” or something. I’m wondering if some are thrown off by the initial name of the PR, and feel that it is implying linting is going to be removed from the default package completely.

sangm:
@webark Done. Thanks for the suggestion 👍

rwjblue:
> how hard is it to add a yarn test or equivalent to the testem config to fail the tests when they are run as normal?

Ya, should be possible assuming we can make the `test` script in package.json essentially be `yarn lint:js && yarn lint:hbs && ember test` in a platform compatible way (I’m not sure the best way to do this that works on both *nix and win).

I do like the idea of `yarn test`/ `npm test` still being the final pass/fail...

webark:
> I do like the idea of yarn test/ npm test still being the final pass/fail...

Oh sure, but i’d say there is a community interest that optionally a lint fail would fail the build. I don’t think they have to be seems as “tests” like the do know, but it would be nice if you could configure testem to run the lint command, then if there where failures, run the tests, and fail.

If testem doesn’t have that functionality.. :/ then that’s too bad, but a feature issue could be raised over there.

sangm:
@rwjblue 

```
  "scripts": {
    "test": "node bin/test-run"
  }
```

```
// bin/test-run.js

// code here to check if window / linux and run correct commands based on platform
```

Seems like the above would work

rwjblue:
Ya, for sure, though I’m hesitant to add another (mostly boilerplate) file to all new apps... 🤔

rwjblue:
Maybe via https://www.npmjs.com/package/npm-run-all?

webark:
i don’t think you have to include a file like that by default.. but a clear description of how one would go about it i think would be enough. even if it was just in this RFC.

sangm:
Should I update the RFC to contain https://github.com/mysticatea/npm-run-all?

Thoughts?

kellyselden:
`&&` is a cross-platform operator, so some kind of `"lint:all": "npm run lint:js && npm run lint:hbs"` should be fine IMO.

rwjblue:
Awesome, that is much simpler/better. I just wasn’t sure it would or wouldn’t work on windows. Thanks @kellyselden!

sangm:
Since this is my first RFC, not sure what the next step is.

1. I can update the RFC with detailed implementation regarding the conversation above (`yarn lint:all`) or even overriding `yarn test`.
2. Wait for further comments from community. (Not sure how long this takes)
3. ??? 

Thanks @rwjblue @kellyselden!

scottkidder:
I have been migrating my apps to this on @rwjblue's advice and it has improved our build times enough that it was worth it just for that.

stefanpenner:
I'm 100% behind what @rwjblue wrote above, +1 on this RFC.

stefanpenner:
Ultimately I think linting as part of tests was a good experiment, but IMHO didn't scale well at all. We can always let folks opt back in, but I don't think it passes the litmus test any longer for "included by default".

We can always include it as part of `ember test` via a custom testem runner, we should consider that?

kellyselden:
+1. Getting our build times under control is worth altering the way people receive their linting errors.

simonihmig:
Regarding the motivation to "Improve our build speed", do we have any numbers what speedup this might yield on average?

Regardless, I am in favor. I guess we should just finish the bike shedding here how the integration should look like, and put that explicitly into the RFC. The options so far seem to be:
1. run standard ESLint as part of `npm lint:js` and have that be run as part of the default Travis config. As Robert pointed out, this is already in place, see https://github.com/ember-cli/ember-new-output/blob/v3.4.1/package.json#L16 and https://github.com/ember-cli/ember-new-output/blob/v3.4.1/.travis.yml#L25-L27. So this just needs the removal of `ember-cli-eslint` in favor of `eslint`, as already mentioned in the RFC.
2. Integrate all tests and linting into `npm test`. This would require rolling back the Travis config to only run `npm test` then, I guess.
3. a custom testem runner

I would tend to 1., given the least amount of friction with existing setups.

Gaurav0:
I think, like ember-welcome-page, we should just make ember-cli-eslint easily removable. It is great for new users and build times for a small app created with ember new are very reasonable. People with large apps who want to optimize should be able to follow instructions to remove it.

I'm afraid core team developers are advanced users and may not be paying attention to the early benefits having linting on by default on each build gives a new user.

rwjblue:
> I think, like ember-welcome-page, we should just make ember-cli-eslint easily removable. 

FWIW - It is already easily removed (`npm uninstall ember-cli-eslint` / `yarn remove ember-cli-eslint`).

Gaurav0:
My build speed for my large app went from 40 s to about 38 s by removing all linting, including eslint, sass, and template. Whatever's making it slow, that isn't it.

rwjblue:
Hmm, I'd personally love a 5% build speed improvement, I'm sorry that doesn't seem important to you.

Gaurav0:
I'm sure it will be important to some. The question is whether we should provide real time eslint feedback in the build vs. a small speed improvement as a default. If someone is getting a massive speed boost out of this, by all means, let them remove ember-cli-eslint.

If we can provide real time eslint feedback without build speed loss, I would be ecstatic. In the meantime, let us discuss the tradeoff.

webark:
@gaurav0 what was the difference in your “test” time?

rwjblue:
@simonihmig:

> Regarding the motivation to "Improve our build speed", do we have any numbers what speedup this might yield on average?

I was also curious just how much time was spent in linting during a fresh app, so I put together a `log-lint-times` package that hooks into ember-cli's `instrumentation` hook and logs out the total linting time for each build / rebuild.

Usage is basically:

```
ember install log-lint-times
EMBER_CLI_INSTRUMENTATION=1 ember s
```

<details>
<summary><b>tldr; Shaves ~ 11% off of `ember new foo` app initial build times (17% off JS rebuild time), and shaves ~ 15% off of medium sized "real world app" build times (~ 5% off JS rebuild time).</b></summary>
<ul>
<li>New addon initial cold build: <pre>Took: 2229.831129ms out of 19170.574034ms</pre></li>
<li> New addon initial warm build: <pre>Took: 1402.544388ms out of 6000.415215ms</pre></li>
<li> New addon JS file rebuild: <pre>Took: 285.099939ms out of 1621.410653ms</pre></li>
<li> New app initial cold build: <pre>Took: 2391.006425ms out of 36832.706141ms</pre></li>
<li> New app initial warm build: <pre>Took: 725.38499ms out of 8341.430844ms</pre></li>
<li> New app JS file rebuild: <pre>Took: 498.726686ms out of 4193.427742ms</pre></li>
<li> <a href="https://github.com/travis-ci/travis-web">travis-web</a> initial cold build: <pre>Took: 45711.571271ms out of 303718.316417ms</pre></li>
<li> <a href="https://github.com/travis-ci/travis-web">travis-web</a> initial warm build: <pre>Took: 15310.792521ms out of 86445.981922ms</pre></li>
<li> <a href="https://github.com/travis-ci/travis-web">travis-web</a> JS file rebuild: <pre>Took: 303.596844ms out of 6510.165714ms</pre></li>
</ul>
</details>

rwjblue:
@Gaurav0:

> The question is whether we should provide real time eslint feedback in the build 

This RFC doesn't remove real time eslint feedback! `ember-cli-eslint` doesn't provide realtime feedback, it provides feedback in the console that is running the server (which is almost _never_ where you actually made a change). If anything it encourages folks to stop using the console output as a crutch and actually install something that **is** realtime.  The only form of feedback in this space that is realtime is something in your editor that shows _while you are typing_ what is wrong.

> vs. a small speed improvement as a default

There are **no silver bullets** the performance game. The improvements that we make are all small and iterative. Over time those gains add up to something substantial.

>If someone is getting a massive speed boost out of this, by all means, let them remove ember-cli-eslint.

Lets be very clear here: this isn't about large apps or small apps. This impacts everyone by a fairly substantial amount as evidenced by my last comment (as well as your own comment above).

jfelchner:
For what it's worth, I didn't know that all that plugin did was add linting during build/test time.  If I'd have known that, I'd have removed it a _long_ time ago.  Not only that, but I've recently had issues with wanting to upgrade ESLint to 5.6 and thinking that I needed to wait for both `ember-cli-eslint` _and_ the Broccoli ESLint addon to get updated in order for that to happen.

Linting is something that my editor does, something that my CI does and something my [git hooks](https://github.com/brigade/overcommit/blob/master/lib/overcommit/hook/pre_commit/es_lint.rb) do.  Linting during tests very quickly becomes a thing that throws me out of my flow.

As long as the default Ember app has CI linting out of the box and a default ESLint config that is sane, I think that'll be good enough for new users.

Gaurav0:
@jfelchner Working on that upgrade for you. https://github.com/ember-cli/broccoli-lint-eslint/pull/135

scottkidder:
Given the lack of consensus here, has adding an option to `ember new` been considered? There is a precedent as `ember new --no-welcome` skips installing ember-welcome-page. Something like `--linting-tests` and `--no-linting-tests` should be simple enough?

OTOH I don't think I have ever used `--no-welcome`, I end up (painlessly) removing `ember-welcome-page` around the time I prepare to deploy the project.

sangm:
@scottkidder that's the problem. I think we are discussing what the "sane default" is. As @jfelchner mentioned, some people aren't even aware this is happening..

That's a good point though. We haven't discussed adding `--lint-test` (by default exclude)

Turbo87:
@sangm every extra option adds quite a bit of complexity to the blueprint code and makes tools like ember-cli-update increasingly complicated. additionally an option is almost as hard to discover for newcomers as the dedicated ember-cli-eslint addon. since installing the addon is only a single ember install away there is not much advantage in having an option but quite a few disadvantages.

rwjblue:
The ember-cli core team has discussed this quite a bit, and are in favor of moving this RFC into final comment period.

Gaurav0:
> The ember-cli core team has discussed this quite a bit, and are in favor of moving this RFC into final comment period.

Was there a meeting? I was hoping to join.

mehulkar:
I understand that removing this package will remove lint assertions from `localhost:4200/tests` and from ember-cli console logs, but I'm not clear if `ember test` will still fail if lint fails. Can we update the RFC to be more clear about this? 

For what it's worth, failing CI builds on lint failures has been much more effective for us than chasing down team members to enable editor plugins. And enforcing pre-commit hooks is a reaally intrusive change, so I'd be against that.

jrjohnson:
@mehulkar `ember test` will not fail, but `yarn lint:js` and `npm run lint:js` *will* fail and those are now part of the default TravisCI blueprint.

rwjblue:
@mehulkar:

> I'm not clear if ember test will still fail if lint fails. Can we update the RFC to be more clear about this?

`npm test` will become `ember test && npm run lint:js && npm run lint:hbs`, and therefore will continue to fail. 

@sangm - Would you mind tweaking to make that clearer?

> For what it's worth, failing CI builds on lint failures has been much more effective for us than chasing down team members to enable editor plugins. 

Yep, totally agree, **however** regardless of this change having your editor setup with nice bit of linting will make for a massively better experience in all JS projects (node, Ember apps, other frameworks, etc...

> enforcing pre-commit hooks is a reaally intrusive change, so I'd be against that.

Me too! I hate precommit hooks 😡

mehulkar:
We don't use Travis, but changing `npm test` is 👍 👍 (this should be added to the RFC)

I *do* currently rely on browser tests to run lint before submitting a PR. But since the goal there is to reduce the number of CI failures, running `npm test` is better anyway (also avoids the whole development environment issue for browser tests). (Or drop down to `./node_modules/bin/eslit app test` since it's way faster.)

💯 for build time improvements and removing unnecessary ember wrapper code. Both are great steps in the right direction for the purpose of marketing Ember, IMO.

> having your editor setup with nice bit of linting will make for a massively better experience in all JS projects

I agree, but realtime linting in VSCode is a HUGE performance drain. I often have to turn off linting to get back responsiveness. But that's both anecdotal and out of scope for this discussion.

Gaurav0:
I've published an ESLint 5 supporting branch of ember-cli-eslint as `@gaurav9576/ember-cli-eslint`, you can test it out now.

sangm:
@rwjblue 

for ember-cli, are we using `npm` or `yarn`?

I ask because

`ember test && npm run lint:js && npm run lint:hbs`

vs

`ember test && yarn lint:js && yarn lint:hbs`

kellyselden:
@sangm We support both.

jnfingerle:
> 
@kellyselden 

> @sangm We support both.

So npm test will be generated as `ember test && npm run lint:js && npm run lint:hbs` or `ember test && yarn lint:js && yarn lint:hbs` depending on the `--yarn` switch at project generation time? Or did I misunderstand that?

kellyselden:
The RFC doesn't specify, but anyone without the --yarn option will get `npm run` commands in their project instead of `yarn`.

jfelchner:
@rwjblue pre-commit hooks are amazing. Fight me. 😂❤️

rwjblue:
@jfelchner - another place, another time 😜

rwjblue:
@sangm - ember-cli supports projects using either, we attempt to detect which is right (based on the `--yarn` flag, presence of `yarn.lock`, etc). For the purposes of the RFC we would basically be doing what @jnfingerle and @kellyselden mentioned (using your package manager of choice).

rwjblue:
After discussing with the rest of the ember-cli team at todays meeting we are all in favor of moving forward. As part of the implementation work we would like to clearly document _how_ to setup `testem.js` to properly run linting (both eslint and template linting) during `ember test --server` and _at least_ link to it from an inline comment in `testem.js` (possibly even include it automatically?).

rwjblue:
Thank you all for the awesome conversation. I truly believe that we came up with a great solution **together**.

